PR #1, opened 2017-10-16
Make API more flexible (specifically, for Outwatch)

raquo:
Changes to support https://github.com/OutWatch/outwatch/pull/62

- Make AttrBuilder, PropBuilder, EventPropBuilder 1) abstract, and 2) not extend any common traits
- Deduplicate the two EventPropBuilder traits
- Add `value` prop
- Remove BoundedBuilder, SpecializedBuilder

Already in master: Remove the need for Modifier type in Styles


shared/src/main/scala/com/raquo/domtypes/generic/builders/EventPropBuilder.scala
```diff
@@ -1,10 +1,10 @@
...
+  * @tparam DomEvent Base type for DOM events (dom.Event in Scala JS)
+  */
+trait EventPropBuilder[EP[Ev <: DomEvent], DomEvent] {
 
-trait EventPropBuilder extends SpecializedBuilder[EventProp] {
-
-  @inline override def build[Ev](key: String): EventProp[Ev] = {
-    new EventProp[Ev](name = key)
-  }
+  @inline def build[Ev <: DomEvent](key: String): EP[Ev]
```
cornerman: Maybe rename `build` to `events` to have similar naming like the other builders?

raquo: Oh definitely, good catch, thanks!

raquo: I think I'll name it `eventProp` though for consistency, and, well – to be truthful – since it canonically returns an `generic.keys.EventProp`, just like the `attr` method canonically returns an `generic.keys.Attr`.

raquo:
I am currently working on resolving naming conflicts between attrs, props, styles and elements when used in one object. I have already compiled a sizeable trait of ReflectedAttrs which could be used to build either props or attrs depending on the implementation of the ReflectedAttrBuilder that it requires. I should have time to finish this within the coming week.

I've also fixed a couple errors in attr typings as part of this upcoming change (one of those was `translate`, which is an enumerated attr that uses "yes" / "no" instead of "true" / "false" because why not? Ugh...)

cornerman:
LGTM, I think this is ready for being merged to master.

For testing, we have setup some very simple [selenium test](https://github.com/woost/wust2/blob/master/systemTest/src/it/scala/WustConnection.scala#L64) with the `PhantomJSDriver` from [seleniumhq](https://github.com/SeleniumHQ/selenium) and specs2 in another project. Maybe helps?

raquo:
Cool, I have a bunch of changes in the pipeline that I want to get out as part of this release, I will push them to this PR some time this week. For example I was able to encode almost all attributes and props as reflected attributes, solving the naming collisions and some other issues.

---

Thanks for selenium examples, I will have to look into it later.

raquo:
@cornerman I have made the changes I wanted, will push everything to this PR tomorrow. Sorry to keep you waiting, very busy week.

cornerman:
@raquo no worries, looking forward to the new version :)

raquo:
So I've pushed my changes. Will break some things, but I gotta fix my mistake.

Notable breaking changes:

**Note:** I've updated the README to explain many of the things here including the why. Classes / traits also have better documentation now.

* Got rid of the concept of BooleanAttr because I think boolean attrs should have the same API in Scala regardless of how they are encoded in HTML & DOM. I decided that I want to abstract away some of that insanity. Instead, there is now a more general concept of Codecs, each of which knows how to encode a Scala value into a DOM / HTML value. It is up to the consuming library to look at the codec that SDT says is applicable to a given attribute and decide how to create the attribute based on that (in case of Outwatch, how to create an attr builder).

* Introduced the concept of Reflected Attributes to eliminate redundancy and dial down the insanity of DOM / HTML API a bit.

* Consolidated some attr and prop traits in `defs` and created new ones

* Applied sane, consistent naming to attribute & props that does not 100% match the DOM API. Only minor things though. Just to make sure that no one needs to google "why setting value attribute does not work". The list of naming differences is in the readme.

* Removed some attr aliases, partly to allow implementers to define their own aliases with good names, partly just to reduce redundancy.

* Fixed incorrect types and names on some attrs and props, removed some props that don't actually exist as props (only attrs), etc.

---

Also:

* Added optional, canonical implementations of key builders (that build our own Attr / Prop / EventProp types)

* Added CompileTest that ensures that an object with all `def` traits on it will compile with no conflicts.

---

I didn't have time to test any of this on my own projects yet, but wanted to push it out ASAP for you to see. I will try to find the time to integrate these changes into Scala Dom Builder & Laminar on the weekend.

raquo:
Hrm `val codec: Codec[V, _]` is causing me grief when using the new API.

When trying to read e.g. an attribute, I am unable to cast the value obtained using `domNode.getAttribute` to this existential type, which is what `decode` expects as input.

I think I'll need to add a DomValue type param to Attr / Prop / ReflectedAttr. Kinda annoying, but seems necessary and will improve type safety.

I'll try to address this tomorrow.

raquo:
@cornerman I can release this as v0.3 if this latest version works for you/Outwatch

raquo:
@cornerman I'm going to release this now as v0.3, let me know if Outwatch compatibility needs anything else

fdietze:
@raquo Thank you very much for your effort! We will give this a try, after our IO-Monad-branch is merged.

PR #3, opened 2017-11-11
make EventProps more usable

cornerman:
This was needed for integrating with outwatch, as we rely on the types of some events that were missing. For `FormEventProps`, I have invented a type parameter `ValueChangeEvent` for events that trigger on value changes in input/textarea/contenteditable (`onInput`, `onChange`, `onSelect`).

This is not complete, as I think, some more events could use more specific typing. But those were the ones we needed so far :)

### DIFF
--- a/js/src/main/scala/com/raquo/domtypes/jsdom/defs/package.scala
+++ b/js/src/main/scala/com/raquo/domtypes/jsdom/defs/package.scala
@@ -6,13 +6,14 @@ import org.scalajs.dom
 package object defs {
 
   object eventProps {
-    type ClipboardEventProps[EP[_ <: dom.Event]] = generic.defs.eventProps.ClipboardEventProps[EP, dom.Event]
+    type ClipboardEventProps[EP[_ <: dom.Event]] = generic.defs.eventProps.ClipboardEventProps[EP, dom.Event, dom.ClipboardEvent]
     type ErrorEventProps[EP[_ <: dom.Event]] = generic.defs.eventProps.ErrorEventProps[EP, dom.Event, dom.ErrorEvent]
-    type FormEventProps[EP[_ <: dom.Event]] = generic.defs.eventProps.FormEventProps[EP, dom.Event]
+    type FormEventProps[EP[_ <: dom.Event]] = FormEventPropsWithValue[EP, dom.Event]
+    type FormEventPropsWithValue[EP[_ <: dom.Event], DomValueElementEvent <: dom.Event] = generic.defs.eventProps.FormEventProps[EP, dom.Event, dom.FocusEvent, DomValueElementEvent]
     type KeyboardEventProps[EP[_ <: dom.Event]] = generic.defs.eventProps.KeyboardEventProps[EP, dom.Event, dom.KeyboardEvent]
     type MediaEventProps[EP[_ <: dom.Event]] = generic.defs.eventProps.MediaEventProps[EP, dom.Event]
     type MiscellaneousEventProps[EP[_ <: dom.Event]] = generic.defs.eventProps.MiscellaneousEventProps[EP, dom.Event]
-    type MouseEventProps[EP[_ <: dom.Event]] = generic.defs.eventProps.MouseEventProps[EP, dom.Event, dom.MouseEvent]
+    type MouseEventProps[EP[_ <: dom.Event]] = generic.defs.eventProps.MouseEventProps[EP, dom.Event, dom.MouseEvent, dom.DragEvent]
     type WindowEventProps[EP[_ <: dom.Event]] = generic.defs.eventProps.WindowEventProps[EP, dom.Event]
   }
 
--- a/js/src/main/scala/com/raquo/domtypes/jsdom/defs/package.scala
+++ b/js/src/main/scala/com/raquo/domtypes/jsdom/defs/package.scala
@@ -2,14 +2,25 @@ package com.raquo.domtypes.jsdom
 
 import com.raquo.domtypes.generic
 import org.scalajs.dom
+import org.scalajs.dom.html
+import scala.scalajs.js.annotation.ScalaJSDefined
 
 package object defs {
+
+  @ScalaJSDefined
+  class ValueElementEvent extends dom.Event {
+    def targetValue: String = target match {
+      case input: html.Input => input.value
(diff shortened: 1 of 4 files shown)


shared/src/main/scala/com/raquo/domtypes/generic/defs/eventProps/ClipboardEventProps.scala
```diff
@@ -7,20 +7,20 @@ import com.raquo.domtypes.generic.builders.EventPropBuilder
   *
   * For type param docs see [[EventPropBuilder]]
   */
-trait ClipboardEventProps[EP[_ <: DomEvent], DomEvent] { this: EventPropBuilder[EP, DomEvent] =>
+trait ClipboardEventProps[EP[_ <: DomEvent], DomEvent, ClipboardEvent <: DomEvent] { this: EventPropBuilder[EP, DomEvent] =>
```
raquo: Pls rename to `DomClipboardEvent` to follow the naming convention (DOM type params start with "Dom")

shared/src/main/scala/com/raquo/domtypes/generic/defs/eventProps/FormEventProps.scala
```diff
@@ -2,37 +2,43 @@ package com.raquo.domtypes.generic.defs.eventProps
 
 import com.raquo.domtypes.generic.builders.EventPropBuilder
 
-trait FormEventProps[EP[_ <: DomEvent], DomEvent] { this: EventPropBuilder[EP, DomEvent] =>
+trait FormEventProps[EP[_ <: DomEvent], DomEvent, DomValueChangeEvent <: DomEvent] { this: EventPropBuilder[EP, DomEvent] =>
```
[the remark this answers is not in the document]
cornerman: That was also my first intuition, but actually these events do not only work on `input` elements: [change-event](https://developer.mozilla.org/en-US/docs/Web/Events/change) (`<input>`, `<select>`, `<textarea>`), [input-event](https://developer.mozilla.org/en-US/docs/Web/Events/input) (`<input>`, `<select>`, `<textarea>`, `contenteditable`), [select-event](https://developer.mozilla.org/en-US/docs/Web/Events/select) (`<input>`, `<textarea>`). So `DomInputElementEvent` might also be confusing, maybe `DomValueElementEvent`?

So we need different handling depending on the element type, i.e., getting the value from element via `value` or `textcontent` property.

shared/src/main/scala/com/raquo/domtypes/generic/defs/eventProps/MouseEventProps.scala
```diff
@@ -24,41 +24,6 @@ trait MouseEventProps[EP[_ <: DomEvent], DomEvent, DomMouseEvent <: DomEvent] {
   lazy val onDblClick: EP[DomMouseEvent] = eventProp("dblclick")
 
   /**
-    * Script to be run when an element is dragged
-    */
-  val onDrag: EP[DomMouseEvent] = eventProp("drag")
```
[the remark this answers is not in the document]
cornerman: done, but moved.

cornerman:
I have adapted the changes according to your comments, except tor the `DomValueElementEvent` - i have added a comment to your now outdated review. Feel free to rename it in every way you want :)

Additionally, I have moved the `onContextmenu` event to the `MouseEventProps`, because it is a mouse event: https://developer.mozilla.org/en-US/docs/Web/Events/contextmenu.

cornerman:
In order to provide a proper implementation of the `DomValueElementEvent` in the `jsdom` package, we would need to adapt [this](https://github.com/OutWatch/outwatch/blob/master/src/main/scala/outwatch/dom/helpers/InputEvent.scala#L9) to also work the value of `input`, `select`, `textarea`, and `contenteditable`.

js/src/main/scala/com/raquo/domtypes/jsdom/defs/package.scala
```diff
@@ -6,13 +6,14 @@ import org.scalajs.dom
 package object defs {
 
   object eventProps {
-    type ClipboardEventProps[EP[_ <: dom.Event]] = generic.defs.eventProps.ClipboardEventProps[EP, dom.Event]
+    type ClipboardEventProps[EP[_ <: dom.Event]] = generic.defs.eventProps.ClipboardEventProps[EP, dom.Event, dom.ClipboardEvent]
     type ErrorEventProps[EP[_ <: dom.Event]] = generic.defs.eventProps.ErrorEventProps[EP, dom.Event, dom.ErrorEvent]
-    type FormEventProps[EP[_ <: dom.Event]] = generic.defs.eventProps.FormEventProps[EP, dom.Event]
+    type FormEventProps[EP[_ <: dom.Event]] = FormEventPropsWithValue[EP, dom.Event]
+    type FormEventPropsWithValue[EP[_ <: dom.Event], DomValueElementEvent <: dom.Event] = generic.defs.eventProps.FormEventProps[EP, dom.Event, dom.FocusEvent, DomValueElementEvent]
```
cornerman: I have added this type alias for convenience.

js/src/main/scala/com/raquo/domtypes/jsdom/defs/package.scala
```diff
@@ -2,14 +2,25 @@ package com.raquo.domtypes.jsdom
...
 
 package object defs {
 
+  @ScalaJSDefined
+  class ValueElementEvent extends dom.Event {
+    def targetValue: String = target match {
+      case input: html.Input => input.value
+      case textarea: html.TextArea => textarea.value
+      case select: html.Select => select.value
+      case elem: html.Element => elem.textContent // for contenteditable elements
```
raquo: Hmmm doesn't look like this should work, have you tried this pattern match in practice with a contentEditable element?

`html.Element` extends `js.Any` (via a few layers), and Scala.js docs [say](https://www.scala-js.org/doc/interoperability/facade-types.html):

> `isInstanceOf[T]` is not supported for any trait `T` inheriting from `js.Any`. Consequently, pattern matching for such types is not supported either.

raquo: Actually wait... `HTMLElement` is an abstract class, not a trait, so I guess it should work

PR #6, opened 2017-11-13
handle events with dom.Element target and access to string value

cornerman:
Another approach on the handling of events with specific targets.

There are more than just `onChange` and `onInput`, so I have added an `ElementTargetEvent`: it has `def target: dom.Element` and exposes a `def targetValue: String` for accessing the text content. This can be used for all events, where we know a more precise type of the `EventTarget`.

For me, this handles most use cases and it is useful to have the `target` as an Element, so you can call methods like `focus`, `blur`, etc. You can also still add your `DomInputEvent`.

What do you think?

### DIFF
--- a/js/src/main/scala/com/raquo/domtypes/jsdom/defs/package.scala
+++ b/js/src/main/scala/com/raquo/domtypes/jsdom/defs/package.scala
@@ -7,18 +7,28 @@ import scala.scalajs.js
 
 package object defs {
 
-  /** Type refinement for events that are fired on input elements.
-    * You can safely use this type for `onChange` and `onSelect` events,
-    * and SOME `onInput` events. See README for more details. */
   @js.native
-  trait InputElementTargetEvent extends dom.Event {
-    override def target: dom.html.Input = js.native
+  //TODO: maybe even ElementTargetEvent[T <: dom.html.Element]?
+  trait ElementTargetEvent extends dom.Event {
+    override def target: dom.html.Element = js.native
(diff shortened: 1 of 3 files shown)


js/src/main/scala/com/raquo/domtypes/jsdom/defs/package.scala
```diff
@@ -7,18 +7,28 @@ import scala.scalajs.js
 
 package object defs {
 
-  /** Type refinement for events that are fired on input elements.
-    * You can safely use this type for `onChange` and `onSelect` events,
-    * and SOME `onInput` events. See README for more details. */
   @js.native
-  trait InputElementTargetEvent extends dom.Event {
-    override def target: dom.html.Input = js.native
+  //TODO: maybe even ElementTargetEvent[T <: dom.html.Element]?
```
cornerman: We could also be more specific, as we know that e.g. `onChange` has target `dom.Input | dom.Textarea' and `onSearch` has `dom.Input`.

raquo: Unfortunately `|` is pretty useless in this case. Even though both`dom.Input` and `dom.Textarea` have a `value` property, you can't just access `target.value` if `target: dom.Input | dom.Textarea`.

raquo:
@cornerman Thanks, this looks good, I've made a few refinements in https://github.com/raquo/scala-dom-types/pull/7

PR #7, opened 2017-11-14
API: FormEventProps and related types

raquo:
- Improve types in FormEventProps
- Move related event trait and class into events package
- Add type param to ElementTargetEvent
- ~~Make ElementTargetEventWithValue non-implicit~~
- Remove ElementTargetEventWithValue


js/src/main/scala/com/raquo/domtypes/jsdom/defs/events/ElementTargetEventWithValue.scala
```diff
@@ -0,0 +1,16 @@
+package com.raquo.domtypes.jsdom.defs.events
+
+import org.scalajs.dom
+
+/** One of the possible ways to access `target.value` on input events.
+  * If you want to use this, define an implicit conversion that instantiates this class.
```
raquo: I mean, the only benefit of keeping that class implicit would have been to let end users import the implicit class from SDT. But I doubt that the consuming library would want end users to be doing that, said library will probably have its own way of managing the implicits it needs.

cornerman: Good point, I agree.

cornerman:
LGTM!

Now proper testing in javascript would be really beneficial to detect potential type problems :)

shared/src/main/scala/com/raquo/domtypes/generic/defs/eventProps/FormEventProps.scala
```diff
@@ -3,76 +3,83 @@ package com.raquo.domtypes.generic.defs.eventProps
...
+  *            This event type has no corresponding type in JS DOM. See our own `ElementTargetEvent` trait.
+  *            - Strictly speaking, this type is not applicable to `onInput` because the latter
+  *              can be fired on any HTMLElement that has `contentEditable` mode enabled.
+  *            - Similarly for `onChange` and `onSelect` – these could also fire on an `HTMLTextAreaElement`.
   * @tparam DomInputEvent
   *            DOM InputEvent https://developer.mozilla.org/en-US/docs/Web/API/InputEvent
   *            Note: This type is not implemented in scala-js-dom
   */
-trait FormEventProps[EP[_ <: DomEvent], DomEvent, DomFocusEvent <: DomEvent, DomInputEvent <: DomEvent, DomElementTargetEvent <: DomEvent] { this: EventPropBuilder[EP, DomEvent] =>
+trait FormEventProps[EP[_ <: DomEvent], DomEvent, DomFocusEvent <: DomEvent, DomInputEvent <: DomEvent, DomHtmlElementTargetEvent <: DomEvent, DomFormElementTargetEvent <: DomEvent, DomInputElementTargetEvent <: DomEvent] { this: EventPropBuilder[EP, DomEvent] =>
```
cornerman: Maybe we could remove the type bounds from the `Dom...ElementTargetEvent`. When we use it we could always do `...Event with Dom...ElementTargetEvent`. Might be better suited for non-jsdom implementation?

js/src/main/scala/com/raquo/domtypes/jsdom/defs/events/ElementTargetEventWithValue.scala
cornerman: Hmm, while using it i feel, this extension might not belong into this library. In the end, I might be going to extend it in the consuming library with things like: `valueAsNumber` or `checked` or go for another concept with something like `def typedTarget[T <: dom.Element] = target.asInstanceOf[T]`. For transparency to other developers it then makes sense to define them in one place where it is visible.

shared/src/main/scala/com/raquo/domtypes/generic/defs/eventProps/FormEventProps.scala
raquo: Hmm I think for now those type constraints are conceptually sound, both in JS and on the JVM. Event handlers should emit events, this is really all that the constraint says. We can revisit this if/when we get a legit use case where this constraint doesn't make sense.

Practically speaking, if we want to use some trait which is unrelated to `DomEvent`, we could just pass `dom.Event with SomethingElse` as a type param to `FormEventProps` instead of having a `with` inside the latter.

PR #9, opened 2017-11-16
More event props improvements

cornerman:


shared/src/main/scala/com/raquo/domtypes/generic/defs/eventProps/ErrorEventProps.scala
```diff
@@ -10,5 +10,5 @@ trait ErrorEventProps[EP[_ <: DomEvent], DomEvent, DomErrorEvent <: DomEvent] {
   /**
     * Script to be run when an error occurs when the file is being loaded
     */
-  lazy val onError: EP[DomErrorEvent] = eventProp("onerror")
```
raquo: Oopsie. Good catch

shared/src/main/scala/com/raquo/domtypes/generic/defs/eventProps/MiscellaneousEventProps.scala
```diff
@@ -7,7 +7,28 @@ import com.raquo.domtypes.generic.builders.EventPropBuilder
...
+    * images and sub-frames have finished loading.
+    *
+    * MDN
+    */
+  lazy val onLoad: EP[DomUIEvent] = eventProp("load")
+
+  /**
+    * Fires when the browser window is resized
+    */
+  lazy val onResize: EP[DomUIEvent] = eventProp("resize")
```
raquo: Why do you want to move `onResize` to this trait? Is it somehow conceptually different from other events in `WindowEventProps`? Looking at [MDN](https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/onresize) this event seems to be window-specific by nature.

---

Looking at `onLoad` and `onScroll`, I see that they can be applied to elements as well. It would be nice to update the comments on those to not be window-specific and to link to https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/onload and https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/onscroll

shared/src/main/scala/com/raquo/domtypes/generic/defs/eventProps/MouseEventProps.scala
```diff
@@ -4,38 +4,41 @@ import com.raquo.domtypes.generic.builders.EventPropBuilder
...
+trait MouseEventProps[EP[_ <: DomEvent], DomEvent, DomMouseEvent <: DomEvent, DomDragEvent <: DomMouseEvent, DomHtmlElementTargetEvent <: DomEvent] { this: EventPropBuilder[EP, DomEvent] =>
 
   /**
     * The click event is raised when the user clicks on an element. The click
     * event will occur after the mousedown and mouseup events.
     *
     * MDN
     */
-  lazy val onClick: EP[DomMouseEvent] = eventProp("click")
+  lazy val onClick: EP[DomMouseEvent with DomHtmlElementTargetEvent] = eventProp("click")
```
raquo: I suspect we can indeed do this, but is there any formal indication that such events will only happen on elements and not e.g. text nodes or SVG elements?

shared/src/main/scala/com/raquo/domtypes/generic/defs/eventProps/MiscellaneousEventProps.scala
cornerman: The reason is that I want to separate event handlers that are exclusive to `Window`, `Document`, `Element` and `HTMLElement`. All three of them implement [`GlobalEventHandlers`](https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers).

But here they differ:
- `Window` additionally implements [`WindowEventHandlers`](https://developer.mozilla.org/en-US/docs/Web/API/WindowEventHandlers) and its own [event handlers](https://developer.mozilla.org/en-US/docs/Web/API/Window#Event_handlers), which more or less corresponds to our `WindowEventProps` (these properties are not available on `Document` or `Element`).
- `Document` also adds some own [event handlers](https://developer.mozilla.org/en-US/docs/Web/API/Document#Event_handlers). We currently do not have a event properties for those.
- `Element` has (two event handler](https://developer.mozilla.org/en-US/docs/Web/API/Element#Handlers) of its own. We currently do not have a event properties for those.
- `HTMLElement` adds the [clipboard and touch events](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement#Event_handlers). The clipboard events are defined in our `ClipboardEventProps`, we do not have touch events.

Maybe, we can also make it more clear somehow, which `EventProps` traits belong to which `EventTarget`?

shared/src/main/scala/com/raquo/domtypes/generic/defs/eventProps/MouseEventProps.scala
cornerman: Good catch, i think you are right! [MDN](https://developer.mozilla.org/en-US/docs/Web/Events/click) actually only says the target type is `Element`.

Then, I was probably wrong to make `onBlur` and `onFocus` in `FormEventProps` be an `HTMLElementTargetEvent`, as they are also only [defined for target `Element`](https://developer.mozilla.org/en-US/docs/Web/Events/blur). Right?

shared/src/main/scala/com/raquo/domtypes/generic/defs/eventProps/MiscellaneousEventProps.scala
cornerman: I will update the comments

shared/src/main/scala/com/raquo/domtypes/generic/defs/eventProps/MouseEventProps.scala
raquo: Yeah I think we should start out being somewhat conservative in this kind of thing, I'd be happy with `dom.Element` where MDN says so. If we start relying on "conveniently incorrect" types the eventual withdrawal will be painful... I'm still hoping that the other approach I mentioned in a comment on this issue in outwatch will work better for this.

js/src/main/scala/com/raquo/domtypes/jsdom/defs/package.scala
```diff
@@ -1,20 +1,26 @@
...
+    type WindowEventProps[EP[_ <: dom.Event]] = generic.defs.eventProps.WindowEventProps[EP, dom.Event, dom.UIEvent, dom.BeforeUnloadEvent, dom.HashChangeEvent, dom.MessageEvent, PageTransitionEvent, dom.PopStateEvent, dom.StorageEvent]
+
+    trait EventPropsInGlobal[EP[_ <: dom.Event]] extends ErrorEventProps[EP] with FormEventProps[EP] with KeyboardEventProps[EP] with MediaEventProps[EP] with MiscellaneousEventProps[EP] with MouseEventProps[EP] { this: generic.builders.EventPropBuilder[EP, dom.Event] => }
+    trait EventPropsInWindow[EP[_ <: dom.Event]] extends EventPropsInGlobal[EP] with WindowEventProps[EP] { this: generic.builders.EventPropBuilder[EP, dom.Event] => }
+    trait EventPropsInDocument[EP[_ <: dom.Event]] extends EventPropsInGlobal[EP] with ClipboardEventProps[EP] { this: generic.builders.EventPropBuilder[EP, dom.Event] => }
+    trait EventPropsInElement[EP[_ <: dom.Event]] extends EventPropsInGlobal[EP] { this: generic.builders.EventPropBuilder[EP, dom.Event] => }
+    trait EventPropsInHTMLElement[EP[_ <: dom.Event]] extends EventPropsInGlobal[EP] with ClipboardEventProps[EP] { this: generic.builders.EventPropBuilder[EP, dom.Event] => }
```
raquo: Hmm... I'm not sure if these new traits should reside in SDT. I don't really see a good way to use more than one of those traits at a time in a typical consuming library. I mean, the names would conflict, and you'd need to instantiate way more eventprop objects than needed... right? Maybe I'm missing something?

I'm guessing #13 is related here

shared/src/main/scala/com/raquo/domtypes/generic/defs/eventProps/MouseEventProps.scala
cornerman: Sounds right, I will update those event props to having target of type `dom.Element`. We can always revisit if there is new information.

js/src/main/scala/com/raquo/domtypes/jsdom/defs/package.scala
cornerman: Actually, I would like sdt to bring those type as they tell me which event properties are applicable to which underlying `EventTarget` type.

My plan is to have the following three usages:
- `trait DomEventProps extends EventPropsInHTMLElement[EmitterBuilder] with EventPropBuilder[EmitterBuilder, dom.Event]` for normal usage within tags as it is now, like: `input(onChange --> sink)`.
- `object WindowEvents extends EventPropsInWindow[Observable] with EventPropBuilder[Observable, dom.Event]` for registering window events via an observable: `WindowEvents.onStorage.map(ev => ...)`.
- `object DocumentEvents extends EventPropsInDocument[Observable] with EventPropBuilder[Observable, dom.Event]` for registering document events via an observable: `DocumentEvents.onLoad.map(ev => ...)`.

The latter two can be build by creating an observable from the callback we have when registering an event with `(Window|Document).attachEventListener(eventKey, callback)`

raquo: Oh that's a nice pattern! I myself was using `node.events(onClick)` but I don't want each node to extend a trait with a hundred vals, even if they're lazy. But for document / window this looks reasonable.

Let's keep all those traits then.

---

I'm assuming you named the new traits `...InX` as opposed to `X...` because `EventPropsInWindow` would otherwise conflict with `WindowEventProps`? Should we maybe rename the original `WindowEventProps` to `WindowOnlyEventProps` or something like that? Would look nicer and would also sort of better match Javascript's stuff:

https://developer.mozilla.org/en-US/docs/Web/API/WindowEventHandlers
https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers

cornerman: Exactly! Sure, we can rename them; I wasn't too happy with this `...In...` :)

cornerman: > node.events(onClick)

With regards to https://github.com/OutWatch/outwatch/issues/93, that approach would work much better with the `currentTarget`-event.

shared/src/main/scala/com/raquo/domtypes/generic/defs/eventProps/FormEventProps.scala
```diff
@@ -19,7 +22,7 @@ import com.raquo.domtypes.generic.builders.EventPropBuilder
   *            DOM InputEvent https://developer.mozilla.org/en-US/docs/Web/API/InputEvent
   *            Note: This type is not implemented in scala-js-dom
   */
-trait FormEventProps[EP[_ <: DomEvent], DomEvent, DomFocusEvent <: DomEvent, DomInputEvent <: DomEvent, DomHtmlElementTargetEvent <: DomEvent, DomFormElementTargetEvent <: DomEvent, DomInputElementTargetEvent <: DomEvent] { this: EventPropBuilder[EP, DomEvent] =>
+trait FormEventProps[EP[_ <: DomEvent], DomEvent, DomFocusEvent <: DomEvent, DomInputEvent <: DomEvent, DomElementTargetEvent <: DomEvent, DomHtmlElementTargetEvent <: DomEvent, DomFormElementTargetEvent <: DomEvent, DomInputElementTargetEvent <: DomEvent] { this: EventPropBuilder[EP, DomEvent] =>
```
cornerman: I am not completely sure about the `target` types anymore. Why does [MDN](https://developer.mozilla.org/en-US/docs/Web/Events/submit) often say `target: Element`, but then in the description gives a more concrete type on which elements the event can occur?

raquo: Well, conventional JS ideology is largely "let's try to run this and see what we get", so I don't imagine JS devs care all that much about what types are specified. There are also potentially some differences in what is specified in the DOM standard and what is actually implemented in browsers, but I'd think those are minimal by now. I think it's mostly just that JS users largely don't care about minor typing details like this.

For the `submit` event MDN says "Note that submit is fired only on the form element, not the button or submit input. (Forms are submitted, not buttons.)", and I've verified this behavior manually in a couple browsers, so I'm confident enough to put `Form` as the type of target for such an event.

raquo:
@cornerman Thanks for working on this! I've just released v0.4

cornerman:
Great! I think, it should be ready now for outwatch. Thank you so much for all your effort!

PR #11, opened 2017-11-16
improve build.sbt

cornerman:
- Due to having different `scalacOptions` for 2.12 and 2.11, they need to be in the setting for `scalaVersion` to resolve correctly for cross version builds
- Added a `inThisBuild` block for build wide configs
- Updated to scala `2.12.4`

Should we make `2.12.4` the default scala version?

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,13 +1,16 @@
+inThisBuild(Seq(
+  name := "Scala DOM Types",
+  normalizedName := "domtypes",
+  organization := "com.raquo",
+  scalaVersion := "2.11.11",
+  crossScalaVersions := Seq("2.11.11", "2.12.4")
+))
+
 // @TODO[WTF] Why can't this be inside releaseSettings?
 releaseCrossBuild := true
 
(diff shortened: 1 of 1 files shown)


raquo:
Not sure about making 2.12.4 the default. In my other projects I use 2.11 because 2.12 has better inference capabilities, and I want to make sure the API is nicely usable from the inferior 2.11 as well.

I assume putting 2.12 helps your development workflow because you use 2.12 for your projects, so you don't need to cross-compile when using a local copy of SDT?

I don't think there are any type inference issues with SDT though, as it's very simple, so I don't really mind making 2.12 the default if it helps somehow.

cornerman:
ok, that makes sense. it is not really a problem, only jitpack somehow does not support cross-builds. But maybe we can soon release a new version? :)

raquo:
I can definitely release a v0.4 if you'd like, just let me know

--
Nikita

cornerman:
I think, a new release would be great after merging the remaining pr.

PR #14, opened 2017-11-24
add `class` alias for className

cornerman:
makes it consistent with `type` which is also there

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,7 +176,6 @@ We try to make the native HTML & DOM API a bit saner to work with in Scala.
   * Note that the `checked` **property** retains its name
 * `selected` **attribute** is renamed to `defaultSelected` for the same reason
   * Note that the `selected` **property** retains its name
-* `class` **attribute** is renamed to `className` for consistency with reflected property name, and to avoid Scala reserved word
 * `for` attribute and `htmlFor` property are available as reflected attribute `forId` for consistency and to avoid Scala reserved word
 * `style` attribute is renamed to `styleAttr` to let you implement a custom `style` attribute if you want.
 * `style` tag is renamed to `styleTag` for the same reason, to avoid hogging a good name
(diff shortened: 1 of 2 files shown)


mariusmuja:
I can't say I agree with the reasoning. There's already more than one option (`cls` and `className`), I don't think adding ` ``class`` `, would create much more confusion. It's available in other popular libraries such as `scalatags` and `scalajs-react`, so there might be more confusion by not having it.
The back-ticks / no back-ticks / what is a better name is subjective.

I don't know what you mean with auto-completion interference, you would just get one more option to autocomplete when you type `clas...`.

In the end it's not a big deal, as you mention, `outwatch` (and other libraries) can include it as an alias.

raquo:
To elaborate on my reasoning:

* `className` exists because that's the actual DOM property name
* `cls` exists as a compromise for brevity and compatibility with ScalaTags/Outwatch code, and because `className` is not a very good name (it usually has a list of class names, not just one class name).

personally I'd be happy including just one of those, but one of the naming design goals I'm following is that end users with knowledge of native DOM names should have no problem finding the right SDT name on their first attempt, with the help of autocompletion. So as the user starts to type intending to write "class" or "className", they will find the "className" SDT name.

On the other hand, the reason I decided to let the key `type` exist is because if you type `type` too quickly, you will miss autocompletion that tells you that `typ` is the right one (the suggestion will not show once you have already typed out `type`).

By autocompletion interference I meant having multiple options available when you type e.g. `clas`. That means you need to either choose or accept whatever IDE gives you as the first option, which is mildly annoying. I often accept autocompletion even before I see what the sluggish IDE has just suggested to me if I know there's only one option. Although I guess nevermind that, I forgot that I already have `classNames` defined instead of `classList` (which I won't use because the native `classList` uses a weird auto-updating JS datastructure).

Regarding scalajs-react, it's worth noting that React.js itself uses the `className` name exclusively. scalajs-react uses its own fork of ScalaTags internally so that's where the `class` name is probably coming from.

cornerman:
I can understand that you want keep the number of aliases at a minimum. But I am not sure whether optimizing for autocompletion should be the only criterion. For example, when working with vim I sometimes do not have autocompletion and then I just guess names or check in the docs/sourcecode of the library. Then I really like it, if the names just match the native html attributes; so backtick names have some intuitive advantage.

But as both of you already said, we can add our own aliases in outwatch, which is also fine.

PR #18, opened 2017-12-13
Default eventProps to simple scala-js-dom types

raquo:
I tried using SDT in a project where I have no use for `TypedTargetEvent`, and it was quite annoying to deal with those types. It was confusing both myself (harder to grasp, somewhat misleading compiler errors) and the IDE (I blame the IDE but still), and required me to write SDT-specific type signatures in end user (non-framework) code instead of simple e.g. `dom.MouseEvent`.

All in all, I don't think it's right to throw this experience onto developers who did not yet face the `target` problem in the first place, nevermind decided to use the SDT solution to it. What's bothering me most is that using those types requires extra knowledge of SDT types in addition to scala-js-dom.

So, I moved the fancy types into a new `typedEventProps` package, and populated the `eventProps` package with basic versions. Migration for Outwatch should consist of just referencing this new package for all the types that are defined in it.

I did change the signatures of a couple `generic` traits as well but it should not matter unless you are referencing those generic traits directly.

I am not a fan of keeping two versions of essentially the same types in SDT, but now I am curious to see which option gets more traction.

cc @cornerman

### DIFF
--- a/js/src/main/scala/com/raquo/domtypes/jsdom/defs/events/PageTransitionEvent.scala
+++ b/js/src/main/scala/com/raquo/domtypes/jsdom/defs/events/PageTransitionEvent.scala
@@ -4,7 +4,7 @@ import org.scalajs.dom
 
 import scala.scalajs.js
 
-//TODO: contribute to scala-js-dom with onpageshow/onpagehide methods
+// @TODO: contribute to scala-js-dom with onpageshow/onpagehide methods
 /** Page transition events fire when a webpage is being loaded or unloaded.
   *
   * See https://developer.mozilla.org/en-US/docs/Web/API/PageTransitionEvent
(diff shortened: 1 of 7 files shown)


cornerman:
Interesting, I am curious what kind of problems you have encountered. So far, I had no issues working with `TypedTargetEvent`, though I did not do too much. But I can totally imagine that error messages can be confusing when there is a type mismatch. What could be done about the IDE issues? Can we somehow improve on the problems?

If this separation is required, I am fine with it and it should be easy to adopt in outwatch. As you said, having now two versions of the same thing is somehow a pity.

raquo:
You know, after looking into this some more, I've realized that I can achieve the same effect by replacing types like `TypedTargetEvent[T] with dom.MouseEvent` with `@js.native trait TypedTargetMouseEvent[T] extends dom.MouseEvent with TypedTargetEvent[T]` together with adding some type bounds on the user side to deal with what is essentially a Subject[Ev] that's invariant in `Ev`.

Now users have a simple `TypedTargetMouseEvent[T]` type that they can easily refer to if needed. It simplifies compiler's "expected type X, required type Y" messages, and doesn't trip IntelliJ anymore. IntelliJ apparently has problems understanding that any instance of `(A extends B) with B` is `B`. Or maybe its type inference is broken, hard to tell without spending too much time there.

So we're back to having just one set of types, with the same package name that was there before.

With the new code, I think for Outwatch no migration steps are needed, but I will still publish it as 0.6 because I'm slightly breaking the API for the underlying `generic` package.

PR #21, opened 2018-01-20
add innerHTML property

cornerman:
This adds the `innerHTML` property, which is useful for feeding raw html into an element. For example, when using a markdown parser and writing the generated html.

### DIFF
--- a/shared/src/main/scala/com/raquo/domtypes/generic/defs/props/Props.scala
+++ b/shared/src/main/scala/com/raquo/domtypes/generic/defs/props/Props.scala
@@ -48,4 +48,11 @@ trait Props[P[_, _]] { this: PropBuilder[P] =>
     * See also [[com.raquo.domtypes.generic.defs.reflectedAttrs.ReflectedAttrs.defaultValue]]
     */
   lazy val value: P[String, String] = stringProp("value")
+
+  /**
+    * The Element.innerHTML property sets or gets the HTML syntax describing the element's descendants.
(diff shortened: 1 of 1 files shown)


raquo:
I... have concerns about this.

`innerHTML` is a dangerous property because whatever you pass to it will be evaluated as HTML, thus making Javascript injection trivial (e.g. XSS attacks). All other attrs and props that we have defined here are safe to pass any value, but not so for innerHTML. That means that you can no longer have a safe function like `setProp[V](prop: Prop[V], value: V)`. If you have such a function you need to be careful to make sure that its params are not coming from the user one way or another, which could be really complicated depending on what your app does.

This can be mitigated by e.g. having that prop accept a special `HTML(val htmlString: String)` type, but I think that's imposing more opinions than SDT should. Some libraries might want to use a different name for this prop (e.g. React uses `dangerouslyInsertInnerHTML`), and other libraries (like Laminar will choose to avoid this prop altogether because the user has access to the underlying DOM node so they could set `innerHTML` manually on it if they really want to.

I think consuming libraries should be defining their own `innerHTML` mechanics.

cornerman:
@raquo Fair point, the usage is indeed unsafe and having a naming or typing scheme to underline this is probably needed. I really like your idea of having something like `div( innerHTML := UnsafeHTML(someString) )`. As we cannot really decided this naming in domtypes, I agree that this belongs into the consuming application/library.

Issue #13, opened 2017-11-17
Track parent tag of properties


cornerman:
Currently, the children properties of a tag do not know, into which kind of tag they are inserted. There is no parent relation of the types.

For example, we can currently write something like `div(value := "bla")`, even though value is only defined on button, option, input, etc and not on div. Additionally, we can currently not type the [`currentTarget`](https://developer.mozilla.org/en-US/docs/Web/API/Event/currentTarget) property of events. This is always the object, on which the event listener was attached. So, if we write `input(onChange := fun)`, we know that the type of `currentTarget` is always `html.Input`.

We would need a sane syntax for writing this. In scala I can only think about a macro for the `tags`, that inject the correct types into the code or adds some implicit.

Also see: https://github.com/OutWatch/outwatch/issues/93

raquo:
Looks like there's two distinct problems here:

P1) Provide type of parent element to keys (attrs / props / etc.)
P2) Restrict type of parent element for each key

Regarding P1, see https://github.com/OutWatch/outwatch/issues/93#issuecomment-345307898 for a possible solution. The idea is that the key itself does not need the type, it's the `:=` method that does, so maybe we can provide the type to it instead.

Regarding P2 – Suppose we could indeed define the type of `value` as `ApplesTo[ParentEl: dom.html.Button | dom.html.Input | dom.html.Option]`. We could then restrict the types provided in P1 to a subtype of `ParentEl`, ensuring that we don't call `value := x` on elements that don't support `value`.

However, the `ParentEl` type from P2 is by itself not very useful – e.g. we can't call `.value` on such a type, and it only works in Scala.js.

So for things like `currentTarget` typing, we would need P1 but not P2.

---

As an alternative to `|` we can have multiple value attrs: `def value: AppliesTo[dom.html.Input]`, `def value: AppliesTo[dom.html.Button]`, etc. but this has its own problems, both for performance (need to instantiate many more objects) and convenience (can't do `if (attr == value)` anymore).

raquo:
Hmm instead of defining `value` as `ApplesTo[ParentEl: dom.html.Button | dom.html.Input | dom.html.Option]` we could define somewhat like this:

Note: I haven't tried to compile any of that. Not sure if it would work.

```
trait ValueAttr

trait AppliesTo[T, P] {
  type Parent = P
}
implicit object valueAppliesToInput AppliesTo[Value, dom.html.Input]
[...]

PR #22, opened 2018-02-07
Svg Support

### LINKED ISSUE #20: SVG support
What needs to be done to support SVG? I'll need this soon.

fdietze:
I was able to port the svg-tags from scalatags via vim-macros!

For reference:
* https://developer.mozilla.org/en-US/docs/Web/SVG
* SVG-Tags in scalatags with MDN documentation:
  https://github.com/lihaoyi/scalatags/blob/44cbd0602d40186d33d39b8d4a259f681d2e4a3b/scalatags/shared/src/main/scala/scalatags/generic/SvgTags.scala
* SVG tag to dom-type association in scalatags
  https://github.com/lihaoyi/scalatags/blob/44cbd0602d40186d33d39b8d4a259f681d2e4a3b/scalatags/js/src/main/scala/scalatags/jsdom/SvgTags.scala

- [ ] Create generic `svg/Attrs.scala` trait with typed listings of SVG attributes
- [x] Same for `svg/Tags.scala` (might need multiple files similar to how HTML tags are done)
- [ ] Add JS-specific type aliases of the above to `jsdom/defs/package.scala`
- [ ] Probably need a new key class: `SvgAttr` (I don't _think_ we should reuse Attr for that), and a canonical builder for it (see `canonical` dir). Don't want to accidentally use SVG attrs on HTML elements, and without #13 not sure what else we can do. Not sure about this, and IIRC you're not using these concrete classes in Outwatch, so I guess I'll need to think about this some time later.
- [ ] For the `Tag` class, I _think_ no need to create `SvgTag`, maybe Tag[SVGElement] would be enough. Not sure.
- [ ] Lastly, we'd need to add SVG to the bundle in CompileTest. I think I'm ok allowing naming collisions between SVG and HTML, since most librarries will probably choose to namespace everything svg-related. Things within SVG should probably be free of collisions and follow the same [naming guidelines](https://github.com/raquo/scala-dom-types#sanity-preservation-measures) as our HTML stuff
- [ ] SVG Namespaces?

fixes #20

### DIFF
--- a/shared/src/main/scala/com/raquo/domtypes/generic/defs/tags/SvgTags.scala
+++ b/shared/src/main/scala/com/raquo/domtypes/generic/defs/tags/SvgTags.scala
@@ -1,0 +1,733 @@
+package com.raquo.domtypes.generic.defs.tags
+
+import com.raquo.domtypes.generic.builders.TagBuilder
+
+trait SvgTags[T[_ <: DomElement], DomElement, SvgElement <: DomElement,
+    SvgCircle <: SvgElement,
+    SvgClipPath <: SvgElement,
+    SvgDefs <: SvgElement,
+    SvgDesc <: SvgElement,
+    SvgEllipse <: SvgElement,
+    SvgFEBlend <: SvgElement,
+    SvgFEColorMatrix <: SvgElement,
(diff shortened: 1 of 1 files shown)


fdietze:
Closing in favor of #23

PR #24, opened 2018-02-09
recreate svg branch

doofin:


shared/src/main/scala/com/raquo/domtypes/generic/defs/attrs/svgAttrs.scala
```diff
@@ -0,0 +1,45 @@
+
+package com.raquo.domtypes.generic.defs.attrs
+
+import com.raquo.domtypes.generic.builders.AttrBuilder
+import com.raquo.domtypes.generic.codecs.{BooleanAsOnOffStringCodec, BooleanAsTrueFalseStringCodec}
+
+/** @tparam A HTML Attribute, canonically [[com.raquo.domtypes.generic.keys.Attr]] */
+trait svgAttrs[A[_]] { this: AttrBuilder[A] =>
```
raquo: traits should be title-cased: SvgAttrs

shared/src/main/scala/com/raquo/domtypes/generic/defs/attrs/svgAttrs.scala
```diff
@@ -0,0 +1,45 @@
+
+package com.raquo.domtypes.generic.defs.attrs
+
+import com.raquo.domtypes.generic.builders.AttrBuilder
+import com.raquo.domtypes.generic.codecs.{BooleanAsOnOffStringCodec, BooleanAsTrueFalseStringCodec}
+
+/** @tparam A HTML Attribute, canonically [[com.raquo.domtypes.generic.keys.Attr]] */
```
raquo: Comment no longer accurate – SVG, not HTML.

Also, we should not reuse the `Attr` key for this because you can't use SVG attrs with HTML elements. We need a new SvgAttr key.

shared/src/main/scala/com/raquo/domtypes/generic/defs/attrs/svgAttrs.scala
```diff
@@ -0,0 +1,45 @@
...
+   * It is frequently useful to define animation as an offset or delta
+   * to an attribute's value, rather than as absolute values. This
+   * attribute said to the animation if their values are added to the
+   * original animated attribute's value.
+   *
+   * Value 	replace | sum
+   *
+   * MDN
+   */
+
```
raquo: Are the rest of SVG attrs TODO at the moment?

shared/src/main/scala/com/raquo/domtypes/generic/defs/tags/SvgTags.scala
```diff
@@ -0,0 +1,733 @@
+package com.raquo.domtypes.generic.defs.tags
+
+import com.raquo.domtypes.generic.builders.TagBuilder
+
+trait SvgTags[T[_ <: DomElement], DomElement, SvgElement <: DomElement,
```
raquo: TODO for myself: Do we need `DomElement` in the type signature at all? Seems like we could get by with just SvgElement

shared/src/main/scala/com/raquo/domtypes/generic/defs/tags/SvgTags.scala
```diff
@@ -0,0 +1,733 @@
...
+  this: TagBuilder[T, SvgElement] =>
+
+
+  /**
+   * The altGlyph element allows sophisticated selection of the glyphs used to
+   * render its child character data.
+   *
+   * MDN
+   */
+  lazy val altGlyph:T[SvgElement] = tag("altGlyph")
```
raquo: formatting: please add a space after `:` for all vals here

shared/src/main/scala/com/raquo/domtypes/generic/defs/tags/SvgTags.scala
```diff
@@ -0,0 +1,733 @@
...
+   * MDN
+   */
+  lazy val clipPath:T[SvgClipPath] = tag("clipPath")
+  
+  /**
+   * The element allows describing the color profile used for the image.
+   *
+   * MDN
+   */
+  lazy val `color-profile`:T[SvgElement] = tag("color-profile")
```
raquo: I think let's expose this as `colorProfile` instead, for consistency with our camelCase naming style.

Same for other backticked vals

shared/src/main/scala/com/raquo/domtypes/generic/defs/tags/SvgTags.scala
```diff
@@ -0,0 +1,733 @@
+package com.raquo.domtypes.generic.defs.tags
+
+import com.raquo.domtypes.generic.builders.TagBuilder
+
+trait SvgTags[T[_ <: DomElement], DomElement, SvgElement <: DomElement,
+    SvgCircle <: SvgElement,
```
raquo: Oof, that's a lot of type params. I looked into a few of those and it seems that they do provide value, thanks for digging those out.

It would be great to batch-rename all these type params to have SvgElement suffix for consistency with other traits. For example: `SvgComponentTransferFunction` -> `ComponentTransferFunctionSvgElement`

shared/src/main/scala/com/raquo/domtypes/generic/defs/attrs/svgAttrs.scala
doofin: Fixed,created new keys/SvgAttr.scala

shared/src/main/scala/com/raquo/domtypes/generic/defs/attrs/svgAttrs.scala
doofin: fixed

shared/src/main/scala/com/raquo/domtypes/generic/defs/attrs/svgAttrs.scala
doofin: Yeah,That would be todo later

PR #25, opened 2018-02-24
 Svg support

### LINKED ISSUE #20: SVG support

doofin:
Seems that we should move every svg thing here,which is idiom github workflow..

For reference:
* https://developer.mozilla.org/en-US/docs/Web/SVG
* SVG-Tags in scalatags with MDN documentation:
  
Scalatags impl:

* SVG tag to dom-type association in scalatags
https://github.com/lihaoyi/scalatags/blob/44cbd0602d40186d33d39b8d4a259f681d2e4a3b/scalatags/shared/src/main/scala/scalatags/generic/SvgTags.scala

* Svg attrs  https://github.com/lihaoyi/scalatags/blob/44cbd0602d40186d33d39b8d4a259f681d2e4a3b/scalatags/shared/src/main/scala/scalatags/generic/SvgAttrs.scala

- [x]  Create generic `svg/Attrs.scala` trait with typed listings of SVG attributes
  (created : https://github.com/raquo/scala-dom-types/blob/svg/shared/src/main/scala/com/raquo/domtypes/generic/defs/attrs/svgAttrs.scala   )
- [x] Same for `svg/Tags.scala` (might need multiple files similar to how HTML tags are done)
  (created: https://github.com/raquo/scala-dom-types/blob/svg/shared/src/main/scala/com/raquo/domtypes/generic/defs/tags/SvgTags.scala )
- [x] Add JS-specific type aliases of the above to `jsdom/defs/package.scala` (https://github.com/raquo/scala-dom-types/blob/svg/js/src/main/scala/com/raquo/domtypes/jsdom/defs/package.scala)
- [x] Probably need a new key class: `SvgAttr` (I don't _think_ we should reuse Attr for that), and a canonical builder for it (see `canonical` dir). Don't want to accidentally use SVG attrs on HTML elements, and without #13 not sure what else we can do. Not sure about this, and IIRC you're not using these concrete classes in Outwatch, so I guess I'll need to think about this some time later.
  (Don't understand this point,Currently I am reusing `lazy val accentHeight = stringAttr("accent-height")`,will there be problems?
)
- [x] For the `Tag` class, I _think_ no need to create `SvgTag`, maybe Tag[SVGElement] would be enough. Not sure.
  (Currently as T[SvgElement])
- [x] Lastly, we'd need to add SVG to the bundle in CompileTest. I think I'm ok allowing naming collisions between SVG and HTML, since most librarries will probably choose to namespace everything svg-related. Things within SVG should probably be free of collisions and follow the same [naming guidelines](https://github.com/raquo/scala-dom-types#sanity-preservation-measures) as our HTML stuff
- [x] SVG Namespaces? (No need to do this in domtypes?)

fixes #20

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,10 @@ Our type definitions are designed for easy integration into any kind of library.
 
 
 
-## Community
+## Community & Contributing
+
+### updating:
+    Currently we are working to add support for svg at svg branch ,help very welcomed!
 
 Please use [github issues](https://github.com/raquo/scala-dom-types/issues) for bugs, feature requests, as well as all kinds of discussions, including questions on usage and integrations. I think this will work better than spreading thin across gitter / stackoverflow / etc. You can _watch_ this project on github to get issue updates if you're interested in following discussions.
 
--- a/README.md
+++ b/README.md
@@ -1,12 +1,12 @@
 # Scala DOM Types
 [![Build Status](https://travis-ci.org/raquo/scala-dom-types.svg?branch=master)](https://travis-ci.org/raquo/scala-dom-types)
 
-This project provides listings and type definitions for Javascript HTML tags as well as their attributes, properties, and CSS styles. 
+_Scala DOM Types_ provides listings and type definitions for Javascript HTML tags as well as their attributes, DOM properties, and CSS styles.
 
     "com.raquo" %%% "domtypes" % "0.5"    // scala.js
     "com.raquo" %% "domtypes" % "0.5"     // JVM
 
-Type definitions in this project are designed for easy integration into any kind of library. You can use this project to build your own DOM libraries like React or Snabbdom, but type-safe. For example, popular Scala.js reactive UI library [Outwatch](https://github.com/OutWatch/outwatch/) recently switched to _Scala DOM Types_, offloading thousands of lines of code and improving type safety ([diff](https://github.com/OutWatch/outwatch/pull/62)). I am also using _Scala DOM Types_ in my own projects:
+Our type definitions are designed for easy integration into any kind of library. You can use this project to build your own DOM libraries like React or Snabbdom, but type-safe. For example, popular Scala.js reactive UI library [Outwatch](https://github.com/OutWatch/outwatch/) recently switched to _Scala DOM Types_, offloading thousands of lines of code and improving type safety ([diff](https://github.com/OutWatch/outwatch/pull/62)). I am also using _Scala DOM Types_ in my own projects:
 
 - [Scala DOM Builder](https://github.com/raquo/scala-dom-builder), a low level DOM manipulation and tree tracking library
 - [Laminar](https://github.com/raquo/laminar), a high level reactive UI library for Scala.js 
@@ -53,8 +53,6 @@ setProperty[Value](element: dom.Element, prop: Prop[Value], propValue: Value)
 
 Now you can't pass just about any random string as `propName`, and even `propValue` is now type checked.
 
-However, if all you want is more type safety in your application code, you might want to import tags / attrs / etc. from [Scala DOM Builder](https://github.com/raquo/scala-dom-builder)'s `simple` package instead because those don't require any boilerplate on your side. 
-
 
 ## What about ScalaTags
 
@@ -62,7 +60,9 @@ However, if all you want is more type safety in your application code, you might
 
 - **More type safe**. For example, in _Scala DOM Types_ an `input` tag is linked to Scala.js `HTMLInputInput` class. This lets you provide exact types for the DOM nodes you create, so that you don't need to perform unsafe casts in your application code if you want to e.g. access the `value` property on an `input` you created. Similarly, all attributes, properties and styles are linked to the types that they accept to prevent you from assigning incorrect values.
 
-- **More flexible**. _Scala DOM Types_ does not tell you how to compose your attributes / props / styles / tags together, and does not enforce any rendering paradigm, you are free to implement your own composition. I see that some projects fork ScalaTags basically to just get the type definitions, I'm hoping there would be no need for that with _Scala DOM Types_.
+- **More flexible**. _Scala DOM Types_ does not tell you how to compose your attributes / props / styles / tags together, and does not enforce any rendering paradigm. You are free to implement your own composition. I see that some projects fork ScalaTags just to get the type definitions without everything else. _Scala DOM Types_ does not get in your way, eliminating the need for such forking.
+
+- **Better representation of native DOM types**. _Scala DOM Types_ handles Reflected Attributes consistently, and uses Codecs to properly encode/decode DOM values.
 
 There are some other differences, for example _Scala DOM Types_ uses camelCase for attr / prop / style names because that is consistent with common Scala style.
 
--- a/README.md
+++ b/README.md
@@ -1,5 +1,6 @@
 # Scala DOM Types
 [![Build Status](https://travis-ci.org/raquo/scala-dom-types.svg?branch=master)](https://travis-ci.org/raquo/scala-dom-types)
+![Maven Central](https://img.shields.io/maven-central/v/com.raquo/domtypes_sjs0.6_2.12.svg)
 
 _Scala DOM Types_ provides listings and type definitions for Javascript HTML tags as well as their attributes, DOM properties, and CSS styles.
 
@@ -13,6 +14,32 @@ Our type definitions are designed for easy integration into any kind of library.
 - [Scala DOM Test Utils](https://github.com/raquo/scala-dom-testutils), a library that verifies that your DOM node / tree matches provided description
 
 `DOM` stands [Document Object Model](https://developer.mozilla.org/en-US/docs/Web/API/Document_Object_Model/Introduction), in our context it's an object that represents an HTML document along with its HTML elements and their attributes, props and styles.
+
+
+
+## Table of Contents
+
+* [Community](#community)
+* [Why use _Scala DOM Types_](#why-use-scala-dom-types)
+* [What about ScalaTags](#what-about-scalatags)
+* [What about scala-js-dom](#what-about-scala-js-dom)
+* [Design Goals](#design-goals)
+* [Documentation](#documentation)
+  * [Codecs](#codecs)
+  * [Reflected Attributes](#reflected-attributes)
+  * [DOM Events & `dom.Event.target`](#dom-events--domeventtarget)
+  * [Naming Differences Compared To Native HTML & DOM](#naming-differences-compared-to-native-html--dom)
+* [Related Projects](#my-related-projects)
+
+
+
+## Community
+
+Please use [github issues](https://github.com/raquo/scala-dom-types/issues) for bugs, feature requests, as well as all kinds of discussions, including questions on usage and integrations. I think this will work better than spreading thin across gitter / stackoverflow / etc. You can _watch_ this project on github to get issue updates if you're interested in following discussions.
+
+See also: [Contribution guide](https://github.com/raquo/scala-dom-types/blob/master/CONTRIBUTING.md)
+
+
 
 ## Why use _Scala DOM Types_
 
@@ -54,6 +81,7 @@ setProperty[Value](element: dom.Element, prop: Prop[Value], propValue: Value)
 Now you can't pass just about any random string as `propName`, and even `propValue` is now type checked.
 
 
+
 ## What about ScalaTags
 
 [ScalaTags](http://www.lihaoyi.com/scalatags/) is a popular Scala library that contains DOM type definitions similar to what we have here. However, _Scala DOM Types_ is different in a few ways:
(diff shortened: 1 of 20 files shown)


README.md
```diff
@@ -33,7 +33,10 @@ Our type definitions are designed for easy integration into any kind of library.
 
 
 
-## Community
+## Community & Contributing
+
+### updating:
+    Currently we are working to add support for svg at svg branch ,help very welcomed!
```
raquo: This is not needed given what's in master and in #26

js/src/main/scala/com/raquo/domtypes/jsdom/defs/package.scala
```diff
@@ -86,6 +86,10 @@ package object defs {
     type EmbedTags[T[_ <: dom.Element]] = generic.defs.tags.EmbedTags[T, dom.Element, dom.html.Image, dom.html.IFrame, dom.html.Embed, dom.html.Object, dom.html.Param, dom.html.Video, dom.html.Audio, dom.html.Source, dom.html.Track, dom.html.Canvas, dom.html.Map, dom.html.Area]
     type TableTags[T[_ <: dom.Element]] = generic.defs.tags.TableTags[T, dom.Element, dom.html.Table, dom.html.TableCaption, dom.html.TableCol, dom.html.TableSection, dom.html.TableRow, dom.html.TableCell, dom.html.TableHeaderCell]
     type MiscTags[T[_ <: dom.Element]] = generic.defs.tags.MiscTags[T, dom.Element, dom.html.Title, dom.html.Style, dom.html.Element, dom.html.Quote, dom.html.Progress, dom.html.Menu]
+    import org.scalajs.dom._
+    import svg._
```
raquo: please don't use those extra imports – instead, spell out `dom.svg.Type` everywhere instead of just `Type` – yes, even in this huge list. It's just much easier to text search, and matches the style in the rest of the definitions here.

js/src/main/scala/com/raquo/domtypes/jsdom/defs/package.scala
```diff
@@ -86,6 +86,10 @@ package object defs {
...
+    type svgTags[T[_ <: dom.Element]]=generic.defs.tags.SvgTags[T,dom.Element,svg.Element,Circle,ClipPath,Defs,Desc,Ellipse,FEBlend,FEColorMatrix,ComponentTransferFunction,FEComposite,FEConvolveMatrix,FEDiffuseLighting,FEDisplacementMap,FEDiffuseLighting,FEFlood,FEFuncA,FEFuncB,FEFuncG,FEFuncR,FEGaussianBlur,FEImage,FEMerge,FEMergeNode,FEMorphology,FEOffset,FEPointLight,FESpecularLighting,FESpecularLighting,FETile,FETurbulence,Filter,G,Image,Line,LinearGradient,Marker,Mask,Metadata,Path,Pattern,Polygon,Polyline,RadialGradient,RectElement,SVG,Stop,SVG,Switch,Symbol,svg.Text,TextPath,TSpan,Use,View]
```
raquo: Pls capitalize the first S so that this type's name is consistent with other type names above

shared/src/main/scala/com/raquo/domtypes/generic/defs/tags/SvgTags.scala
```diff
@@ -0,0 +1,733 @@
...
+    SvgMetadata <: SvgElement,
+    SvgPath <: SvgElement,
+    SvgPattern <: SvgElement,
+    SvgPolygon <: SvgElement,
+    SvgPolyline <: SvgElement,
+    SvgRadialGradient <: SvgElement,
+    SvgRectElement <: SvgElement,
+    Svg <: SvgElement,
+    SvgStop <: SvgElement,
+    SvgSVG <: SvgElement,
```
raquo: This type param seems to be unused? There's another Svg two lines above this

shared/src/main/scala/com/raquo/domtypes/generic/defs/tags/SvgTags.scala
```diff
@@ -0,0 +1,733 @@
+package com.raquo.domtypes.generic.defs.tags
+
+import com.raquo.domtypes.generic.builders.TagBuilder
+
+trait SvgTags[T[_ <: DomElement], DomElement, SvgElement <: DomElement,
```
raquo: `T[_ <: DomElement]` should be `T[_ <: SvgElement]`, I think, and it doesn't seem like we need the `DomElement` type param at all.

shared/src/main/scala/com/raquo/domtypes/generic/defs/tags/SvgTags.scala
```diff
@@ -0,0 +1,733 @@
...
+  this: TagBuilder[T, SvgElement] =>
+
+
+  /**
+   * The altGlyph element allows sophisticated selection of the glyphs used to
+   * render its child character data.
+   *
+   * MDN
+   */
+  lazy val altGlyph:T[SvgElement] = tag("altGlyph")
```
raquo: Formatting: Pls add a space between `:` and `T` in all these lazy vals to be consistent with other traits like this

shared/src/main/scala/com/raquo/domtypes/generic/defs/tags/SvgTags.scala
```diff
@@ -0,0 +1,733 @@
...
+   */
+  lazy val metadata:T[SvgMetadata] = tag("metadata")
+  
+  /**
+   * The missing-glyph's content is rendered, if for a given character the font
+   * doesn't define an appropriate glyph.
+   *
+   * MDN
+   */
+  lazy val `missing-glyph`:T[SvgElement] = tag("missing-glyph")
```
raquo: We should not be using backticks. Tags with dashed should be camelCased instead. (same for other tags)

shared/src/main/scala/com/raquo/domtypes/generic/keys/SvgAttr.scala
```diff
@@ -0,0 +1,10 @@
+package com.raquo.domtypes.generic.keys
+
+import com.raquo.domtypes.generic.codecs.Codec
+
+/**
+  * This class represents an HTML Element Attribute. Meaning the key that can be set, not the whole a key-value pair.
```
raquo: Update "HTML" to "SVG"

shared/src/main/scala/com/raquo/domtypes/generic/defs/attrs/svgAttrs.scala
```diff
@@ -0,0 +1,45 @@
+
+package com.raquo.domtypes.generic.defs.attrs
+
+import com.raquo.domtypes.generic.builders.AttrBuilder
+import com.raquo.domtypes.generic.codecs.{BooleanAsOnOffStringCodec, BooleanAsTrueFalseStringCodec}
+
+/** @tparam A svg Attribute, canonically [[com.raquo.domtypes.generic.keys.SvgAttr]] */
```
raquo: Formatiing: svg should be capitalized ("SVG") in the comment

js/src/main/scala/com/raquo/domtypes/jsdom/defs/package.scala
raquo: Formatting: add spaces around `=`, and put each type param on a new line

shared/src/main/scala/com/raquo/domtypes/generic/defs/tags/SvgTags.scala
doofin: That's true,gonna do that later with regex after initial test

doofin:
CI gives error due to that some attr name is the same as element name ..

[error] /home/travis/build/raquo/scala-dom-types/js/src/test/scala/com/raquo/domtypes/CompileTest.scala:76: object svg inherits conflicting members:
[error]   lazy value clipPath in trait SvgAttrs of type com.raquo.domtypes.generic.keys.SvgAttr[String]  and
[error]   lazy value clipPath in trait SvgTags of type com.raquo.domtypes.generic.builders.Tag[org.scalajs.dom.svg.ClipPath]
[error] (Note: this can be resolved by declaring an override in object svg.);
[error]  other members with override errors are: cursor, filter, mask
[error]     object svg
[error]            ^
[error] one error found
[error] (domtypesJS/test:compileIncremental) Compilation failed
[error] Total time: 19 s, completed Mar 1, 2018 4:29:50 AM

Should we namespace them ? Or provide alternative name?

raquo:
Yep, that's exactly what CompileTest is for, nice.

Unless we know that the `clipPath` attr is much more popular (more often used) than the `clipPath` tag (or the other way around), we should disambiguate them by renaming them `clipPathAttr` and `clipPathTag` respectively.

BTW you don't need to wait for CI, you can run the tests locally with `sbt test`.

doofin:
I have added postfix "Attr" for the attrs and now it passed.Thank for the advice!

Maybe it's time to use dom builder to test svg rendering?
https://github.com/raquo/scala-dom-builder/pull/3

raquo:
For DOM Builder work, we'll need a sample SVG file I described in this comment: https://github.com/raquo/scala-dom-types/issues/20#issuecomment-368217967

doofin:
Do you mean something like `svg(circle(rx:=5,ry:=5))` ? But this syntax seems only available at dom builder which is yet to be finished..

raquo:
No I mean we need such example in plain natural SVG (that is <svg blah=""
etc.>). It should contain attributes of all possible different datatypes
(e.g. numbers, strings, booleans)

--
Nikita

doofin:
So you want to make dom builder render to string and test their equality,right? Where should i put these svg file?

doofin:
Plenty of thanks to lihaoyi (http://www.lihaoyi.com/scalatags/#scalatags.TextSVG)
`<svg height="800" width="500"><polyline points="20,20 40,25 60,40 80,120 120,140 200,180" fill="none" stroke="black" stroke-width="3"></polyline><line x1="175" y1="100" x2="275" y2="0" stroke="rgb(255,0,0)" stroke-width="10"></line><rect x="300" y="10" rx="20" ry="20" width="100" height="100" fill="rgb(0,0,255)" stroke-width="3" stroke="rgb(0,0,0)" fill-opacity="0.1" stroke-opacity="0.5"></rect><circle cx="30" cy="250" r="10" stroke="black" stroke-width="3" fill="red"></circle><ellipse cx="150" cy="250" rx="100" ry="50" fill="yellow" stroke="purple" stroke-width="4"></ellipse><polygon points="300,110 350,290 260,310" fill="line" stroke="purple" stroke-width="10"></polygon><path d="M100 300 L25 500 L175 500 Z"></path><text x="350" y="250" fill="red" transform="rotate(30 20, 40)">I love SVG!</text><text x="350" y="350" fill="green">Several lines<tspan x="350" y="380">First line.</tspan><tspan x="350" y="410">Second line.</tspan></text><defs><linearGradient id="grad1" x1="0%" y1="0%" x2="100%" y2="0%"><stop offset="0%" stop-color="rgb(255,255,0)"></stop><stop offset="100%" stop-color="rgb(255,0,0)"></stop></linearGradient></defs><ellipse cx="100" cy="590" rx="85" ry="55" fill="url(#grad1)"></ellipse></svg>`

raquo:
Cool, looks like it should work. I'll try to get something going on DOM Builder side on the weekend. Sorry for the delay, very busy days now between work and various chores.

doofin:
No worry,after all we are doing this for fun.I would also become very busy soon due to that I got a great scala job about big data . Scala is quite popular these days in China,which is unimaginable!

raquo:
@doofin I almost finished integrating SVG into DOM Builder and Laminar. While I'm wrapping that up, could you please copy over ScalaTags/MDN comments to remaining SVG attrs?

Make sure to first pull the latest changes from this branch to avoid merge conflicts, as I changed the SvgAttrs file among others.

doofin:
@raquo which svg attrs are missing? I think they are all here now 

https://github.com/raquo/scala-dom-types/blob/master/shared/src/main/scala/com/raquo/domtypes/generic/defs/attrs/SvgAttrs.scala

doofin:
It's not present in scalaTags either..so I would have to get them from MDN site myself.I would be happy to write a parser to get them from web automatically  but  currently I am too busy to do that

raquo:
Ah I see, I didn't know they were missing from ScalaTags too. Definitely no
need to spend more time on that then. I'll keep #29 up and if anyone cares
enough they can do it.

Thanks for your help with SVG! I will publish v0.6 on the weekend. SDB and
Laminar stuff I already put in PRs, will be merging all that hopefully this
weekend too.

—
Best regards,
Nikita

fdietze:
Thank you all for working on this PR!

doofin:
Cheers ! Finally we can do some very complex stuff with svg and reactive programming!

PR #30, opened 2018-04-10
Build: Add jitpack.yml for cross-build publish

cornerman:
Otherwise, jitpack will only publish for 2.12.


raquo:
Thanks, I don't use jitpack myself, so if something doesn't work someone will need to tell me.

cornerman:
Sure. We are using jitpack quite heavily - so I will keep an eye on it :)

PR #31, opened 2018-04-14
API: Better separate HTML/SVG tags and attributes

raquo:
This reduces the amount of boilerplate needed to implement your own builders because you can now make a single trait to extend multiple DOM Types traits. Previously, inheritance conflicts between HTML and SVG related traits prevented using this pattern to its full extent.

This is a breaking change that I will release shortly as v0.7


shared/src/main/scala/com/raquo/domtypes/generic/defs/attrs/Attrs.scala
```diff
@@ -1,18 +1,18 @@
 package com.raquo.domtypes.generic.defs.attrs
 
-import com.raquo.domtypes.generic.builders.AttrBuilder
+import com.raquo.domtypes.generic.builders.HtmlAttrBuilder
 import com.raquo.domtypes.generic.codecs.{BooleanAsOnOffStringCodec, BooleanAsTrueFalseStringCodec}
 
-/** @tparam A HTML Attribute, canonically [[com.raquo.domtypes.generic.keys.Attr]] */
-trait Attrs[A[_]] { this: AttrBuilder[A] =>
+/** @tparam A HTML Attribute, canonically [[com.raquo.domtypes.generic.keys.HtmlAttr]] */
+trait Attrs[A[_]] { this: HtmlAttrBuilder[A] =>
```
cornerman: @raquo Maybe would be more consistent to rename Attrs to `HtmlAttrs` here? That would be same as with SvgAttrs and AriaAttrs.

PR #33, opened 2018-06-22
make autocomplete attribute accept string instead of boolean

cornerman:
Just stumbled over the `autocomplete` attribute, which currently just accepts boolean, which are converted to "off/on" values. This is correct for form elements, but input elements accept many more options.

I could not find a proper documentation, but possible values are listed [on mdn](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input) (search for autocomplete). Maybe also relevant: https://www.chromium.org/developers/design-documents/create-amazing-password-forms.


raquo:
Good catch, thanks! I'll get this released as 0.7.1 in the coming days (as soon as I have a moment)

PR #38, opened 2018-07-30
[WIP] Add grid styles

cornerman:
I started adding grid styles but according to MDN they are still a candidate or a working-draft. So I am not sure whether they belong into sdt. Or can we maybe add something like an `experimental` package to mark unstable APIs? This is what scalajs does too for some experimental APIs in the browser (for example serviceworker).

based on #37


cornerman:
Agreed that we should put them in a separate trait. We can make this trait generic on the different value-types we have in the grid directives. Then consuming libraries and applications can implement them in a more type-safe way. We can just default to String for now and improve if we have time and need for it :)

Will update the PR soonish!

jberggg:
Stumbled across this PR when I was trying to find out how to apply grid styles :-)

Should I try to move those changes in a separate trait and make the CI run?

raquo:
So uh sorry I didn't actually explicitly answer your question (in case it was directed at me): you're welcome to adapt this PR (yes, a separate trait for grid-specific props would be best), but you should be aware that it will take a long time for the changes to actually be released / become usable from Laminar.

jberggg:
Hey @raquo 

Ah good to know. Thank you for the update. I'm playing around with Outwatch for a personal project. So this addition is not urgent at all and I'm aware that you can "manually" set CSS properties.  When I find free time I will have a look at the new encoding style and might give it a try. It's not top priority so please do not expect too much 😃

raquo:
Scala DOM Types has been refactored and its structure is now completely different, so I'm closing this old PR. If anyone wants to take another shot at this in the future, please open a new issue or PR.

PR #39, opened 2018-08-01
add missing svg attrs (refX,orient)

doofin:
not sure why some attrs are missing ,here is a test

```scala

import com.raquo.laminar.api.L.svg._
object svgs {
  defs(
    marker(
      id := "",
      viewBox := "0 -5 10 10",
      refX := "4",
      markerWidth := "3",
      markerHeight := "3",
      orient := "auto",
      path(
        d := "M0,-5L10,0L0,5"
      )
    )
  )
}

```


raquo:
Thanks!

For the future, please don't use W3C spec text in our comments, because I'm too lazy to comply with the W3C document license, it requires more detailed copyright notices than MDN. I changed the docs to use MDN information instead. If W3C information is required above what's provided in MDN we can just link to the relevant W3C page instead.

doofin:
@raquo  You are wecome! I would pay attention to that in the future.

PR #40, opened 2018-08-02
more doc

doofin:


cornerman:
Since I know the code by now, I do not need it myself. But it would probably help newcomers to understand the package structure.

There is the generic package which defines the attributes, properties, and events in a platform-independent way (including sameRefTags for a simple integration). Then there is the jsdom package binding the generic definitions to types from the dom. I think this distinction is quite important to get, in order to use and extend this library.

raquo:
I've added a few paragraphs of explanation based on a practical Laminar example, that'll have to suffice for now. Further specific questions can have their own issues.

PR #42, opened 2018-08-17
New: Add `namespace` param to `SvgAttr`

raquo:
For https://github.com/raquo/Laminar/issues/31

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,11 @@ The following allowances for breaking changes exist _for now_:
 
 TODO: Adopt more mature versioning, eventually. Read about eviction, binary compatibility, etc.
 
+#### v0.8 – TBD
+
+* **New: Add `namespace` param to `SvgAttr`**
+* **Naming: `xlink` -> `xlinkRole` to match native attribute name**
+
 #### v0.7.1 – Jul 2018
 
 * **Fix: `autocomplete` is now a string attribute ([#33](https://github.com/raquo/scala-dom-types/pull/33))**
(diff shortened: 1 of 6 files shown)


cornerman:
Good point. In the laminar issue, you mention that e.g. `xlinkHref` needs a namespace which is not `null`. Shouldn't we add it then [here](https://github.com/raquo/scala-dom-types/blob/12446e4cb8782110aea57320c37271e3025ba111/shared/src/main/scala/com/raquo/domtypes/generic/defs/attrs/SvgAttrs.scala#L1444) in this PR?

raquo:
Good catch, yes, that was my intent, will do

--
Nikita

doofin:
@cornerman  is that related to https://github.com/raquo/scala-dom-builder/pull/5 ?

fdietze:
Here are good introductions to xml namespaces and namespace prefixes:
https://developer.mozilla.org/en-US/docs/Web/SVG/Namespaces_Crash_Course
https://www.w3schools.com/xml/xml_namespaces.asp

While this PR looks good to me, I'm still unsure if baking the `xlink` namespace into `scala-dom-types` makes sense. I'm not experienced enough with xml namespaces to decide this.

raquo:
I think it definitely makes sense from a purely _practical_ standpoint – SVG `xlink` attributes are supposedly popular, and our implementation seems to be the universally accepted way of representing them. I mean, I don't think anyone would _reasonably_ want to associate the `xlink` prefix with a different namespace string, even if technically that's possible in the spec.

I might consider extracting xlink attributes into a separate trait in future releases if a case is made for that, but I don't really see the point right now.

Issue #41, opened 2018-08-09
Unescaped text entity -- raw()


marklister:
Do you have an equivalent to scalatags' raw()?

marklister:
The use case is explained in https://www.lihaoyi.com/scalatags/#Auto-escapingandunsanitizedInput

Basically if you want a text entity rendered without escaping "<" ">" etc in scalatags you enclose them in `raw()`.  I was hoping I could do the same in scala-dom-types

raquo:
Hi Mark, my current policy is to not include such unsafe constructs in
scala dom types. SDT is designed to be consumed by higher level libraries,
each of which might have different opinions on how to handle these cases.
For example, in Lamjnar my policy is "the library will not provide this,
use native DOM methods".

In SDT terms, you can implement the `raw` tag or `innerHtml` property in
your own project similarly to how other tags and properties are implemented.

Alteratively, you can simply implement a `def raw(u safeHtmlContent:
String): Modifier[dom.Element]` method and expose it to your code/users.
The method would return a Modifier that sets the given element's
`innerHtml` property to the provided HTML string.

--
Nikita

cornerman:
@raquo How about introducing an `UnsafeAttributes` trait? A consuming library can decide whether it wants to pull this in and if does not want to, the developer can just ignore it. It can also decide to make these definitions only accessible via an `unsafe`-prefix like `unsafe.innerHTML`.

I think, SDT is for typing html/svg/css elements and even if we think `innerHTML` might not be desired by default or is unsafe to do, it is part of the spec after all. It even might be totally fine in some consuming libraries. In outwatch (with a patched version of snabbdom :/), we can work with `innerHTML`. And it is useful if I want to e.g. render markdown as html.

raquo:
Hrm, I guess we can do it this way for `innerHtml`, with the trait having a generic param for the types of values that this property should accept (e.g. not just string, but let the consuming libraries require some `RawHtml(string)` as input). Maybe this will do a better job of informing potential consuming libraries of the dangers of `innerHtml` than omitting it. Although, that would require the user to implement the `innerHtml` method since we won't be able to provide a codec, which... leaves no usefulness to the trait other than the single `innerHtml` name.

I'm not sure how ScalaTags' `raw` works, but it probably _appends_ to the parent's content rather than _replaces_ it like `innerHtml` does. That is definitely out of scope for SDT.

Related: https://github.com/raquo/scala-dom-types/pull/21

[...]

PR #43, opened 2018-08-26
New: Complex Keys (cls, rel, textContent, etc.)

### LINKED ISSUE #32: Make it possible to have a different API for cls / className / classNames
As described in https://github.com/raquo/Laminar/issues/22, consuming libraries might want to provide a special way to deal with CSS classes which is different from how they deal with other  attributes.

It would benefit the consuming libraries if they could choose whether to keep the default cls / className / classNames API, but currently it's not possible.

I guess the most obvious solution to this problem would be to extract css / className / classNames (and perhaps also styleAttr?) into a separate OpinionatedAttrs trait that the consuming library may choose to not use.

raquo:
This PR provides a way for consuming libraries to define custom interfaces for some composite or otherwise special keys.

The attributes I chose are: `cls`, `className`, `rel`, `dataAttr`, `styleAttr`, `textContent`.

I was thinking of also adding `role` attribute to the list – it's also a composite attribute (space-separated strings), but I've never used it and I'm not sure what kind of usage patterns it has. Does it even make sense to have special APIs for it...

Fixes #32.

# TODO
- [x] Change v0.9 date from Aug to Sep

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,8 +11,9 @@ TODO: Adopt more mature versioning, eventually. Read about eviction, binary comp
 #### v0.9 – Sep 2018
 
 * **API: `AriaAttrs` now exposes attributes as members, without `aria` object**
-* **API: Move `cls`, `className`, `rel`, `styleAttr`, and `textContent` into new `ComplexHtmlKeys` and `ComplexSvgKeys` traits, allowing for more freedom to define these keys in non-standard ways. Use `CanonicalComplexHtmlKeys` and `CanonicalComplexSvgKeys` to retain v0.8 functionality**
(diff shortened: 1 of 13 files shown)


raquo:
Agreed, I think I'll change `className` to be an alias for `cls`. Can't really imagine libraries wanting those exact names but different implementations.

I think I'll add the `role` attribute in there after all, for the sake of consistency. 

What do you think about removing `textContent` from this trait (and SDT) for reasons I mentioned in #41?

> In contrast to these fringe properties, ComplexHtmlKeys is supposed to have keys that every library would want to implement, but perhaps differently. TBH I'm still second guessing even adding textContent to it, I might actually remove it from there because I'm not sure how universally supported its behaviour – modifying the tree of elements – would be in consuming libraries.

cornerman:
Yep, then better remove `textContent` as it falls into the same class as `innerHTML`

PR #49, opened 2019-01-20
PointerEventProps

Busti:
Pointer events are now available [across all major browsers](https://caniuse.com/#feat=pointer-events), since they provide [several advantages](https://developers.google.com/web/updates/2016/10/pointer-events) over traditional mouse and touch events, it would be handy to be able to use them in scala-dom-types.

### DIFF
--- a/js/src/main/scala/com/raquo/domtypes/jsdom/defs/package.scala
+++ b/js/src/main/scala/com/raquo/domtypes/jsdom/defs/package.scala
@@ -41,7 +41,7 @@ package object defs {
       EP,
       dom.Event,
       dom.PointerEvent,
-      TypedTargetPointerEvent[dom.Element],
+      TypedTargetPointerEvent[dom.Element]
     ]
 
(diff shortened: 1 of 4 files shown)


js/src/main/scala/com/raquo/domtypes/jsdom/defs/package.scala
```diff
@@ -37,6 +37,13 @@ package object defs {
       dom.DragEvent
     ]
 
+    type PointerEventProps[EP[_ <: dom.Event]] = generic.defs.eventProps.PointerEventProps[
+      EP,
+      dom.Event,
+      dom.PointerEvent,
+      TypedTargetPointerEvent[dom.Element],
```
raquo: This trailing comma is failing Scala 2.11 build

Busti: I fixed the trailing comma. But it looks like i messed up the rebase.  
I can re-create the PR if you wish.

PR #52, opened 2019-08-22
Add scala 2.13

zakpatterson:


build.sbt
```diff
@@ -2,8 +2,8 @@ inThisBuild(Seq(
   name := "Scala DOM Types",
   normalizedName := "domtypes",
   organization := "com.raquo",
-  scalaVersion := "2.12.9",
-  crossScalaVersions := Seq("2.11.12", "2.12.9")
+  scalaVersion := "2.12.8",
```
fdietze:
```suggestion
  scalaVersion := "2.12.9",
```

zakpatterson: Ah yeah, I keep switching that value to 2.12.8 for metals.

fdietze: haha, same for me :wink:, metals has 2.12.9 support now on master. You can use the latest snapshot.

build.sbt
```diff
@@ -45,24 +45,27 @@ lazy val scalacSettings = Seq(
...
+        "-Ywarn-value-discard" ::
+        "-Ywarn-nullary-override" ::
+        "-Ywarn-nullary-unit" ::
+        "-Ywarn-unused" ::
+        "-Xfuture" ::
         Nil
-      case _ =>
+      case Some((2, minor)) if minor < 12 =>
+        "-Ypartial-unification" :: 
+        "-Xfuture" ::
```
raquo: Thanks for the PR @zakpatterson! Question: why are all these scalac options conditional based on Scala version now? At least `-Ywarn-unused` is surely supported by all relevant Scala versions.

zakpatterson: Yes you are right. I added a few more options to apply to all versions. I'm not sure what the best policy is. I usually do fatal warnings on 2.12 only so that the CI blocks all of them, and say that's good enough. If you'd like I can do something comprehensive similar to this in another PR: https://github.com/OutWatch/outwatch/pull/336

raquo: I would prefer to keep compiler options the same for all scala versions, except as needed. So everything that you removed from `scalacSettings` should still apply to 2.11 and 2.12 like it did before your change (currently you only apply most of those to 2.12, not 2.11).

Scala 2.13 config should ideally _behave_ similarly to 2.11 and 2.12 config, but Scala 2.13 either removed or deprecated some of these options (e.g. `-Xfuture`, `-Ypartial-unification`) so I guess those should be omitted from our 2.13 config, but most other options are still supported (or sometimes renamed) and should be included.

raquo: To clarify, warnings are not fatal in our CI config. I'm not opposed to enabling that, but that's a separate issue that shouldn't affect consistency of compiler options across scala versions in my view.

zakpatterson: @raquo I redid this in 0cde935, basically taking the original set of settings, and subtracting the ones that don't work in 2.13. So it's as you say, the same as before.

raquo:
Released as 0.95 to Maven, the central repo will update soon.

PR #54, opened 2019-08-23
WIP: inline def

### LINKED ISSUE #53: Discussion about lazy val definitions
Currently, domtypes defines dom attributes, properties, styles, events and elements as `lazy val`.

Now, we use domtypes in OutWatch, where most of the things we are constructing are builders. For example, the attribute `contenteditable` is an `AttributeBuilder` to which you can assign `Boolean` values. So you would write something like `contenteditable := true`. This is evaluated into `Attribute("contenteditable", "true")`. Ideally, our code should be compiled into the evaluated Attribute directly. Now, our `AttributeBuilder` already uses `inline` annotations to inline most of the construction part, but the compiler cannot inline the whole expression, because `contenteditable` is defined as a lazy val.

I would like to evaluate whether we can use `@inline def` instead of `lazy val` for the definitions in domtypes. What do you think? What kind of tests/evaluations would you like to see?

fdietze:
fixes #53

I had this branch lying around...

### DIFF
--- a/shared/src/main/scala/com/raquo/domtypes/generic/builders/Tag.scala
+++ b/shared/src/main/scala/com/raquo/domtypes/generic/builders/Tag.scala
@@ -1,7 +1,7 @@
 package com.raquo.domtypes.generic.builders
 
 /** Tag represents an Element builder */
-class Tag[+Element](
+@inline class Tag[+Element](
   val name: String,
   val void: Boolean
 )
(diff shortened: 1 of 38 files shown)


raquo:
All of this is now configurable in the new code generator, so no code changes are required in SDT unless we experiment and find a better default. Therefore, closing.

PR #60, opened 2020-12-23
Support animation events

nghuuphuoc:
should close https://github.com/raquo/Laminar/issues/74

### DIFF
--- a/shared/src/main/scala/com/raquo/domtypes/generic/defs/eventProps/MiscellaneousEventProps.scala
+++ b/shared/src/main/scala/com/raquo/domtypes/generic/defs/eventProps/MiscellaneousEventProps.scala
@@ -7,7 +7,44 @@ import com.raquo.domtypes.generic.builders.EventPropBuilder
   *
   * For type param docs see [[EventPropBuilder]]
   */
-trait MiscellaneousEventProps[EP[_ <: DomEvent], DomEvent, DomUIEvent <: DomEvent] { this: EventPropBuilder[EP, DomEvent] =>
+trait MiscellaneousEventProps[
+  EP[_ <: DomEvent],
+  DomEvent,
+  DomAnimationEvent <: DomEvent,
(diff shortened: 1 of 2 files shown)


shared/src/main/scala/com/raquo/domtypes/generic/defs/eventProps/MiscellaneousEventProps.scala
```diff
@@ -7,7 +7,44 @@ import com.raquo.domtypes.generic.builders.EventPropBuilder
...
+] { this: EventPropBuilder[EP, DomEvent] =>
+  /**
+    * An animationcancel event is sent when a CSS animation unexpectedly aborts, that is, any time it stops running
+    * without sending an animationend event. This can happen, for example, when the animation-name is changed such
+    * that the animation is removed, or when the animating node is hidden—either directly or because any of its
+    * containing nodes are hidden—using CSS.
+    *
+    * MDN
+    */
+  lazy val onAnimationCancel: EP[DomAnimationEvent] = eventProp("onanimationcancel")
```
raquo, line 24: Nice, thanks for the PR!

Just one thing: the "onanimationcancel" string here should not have the "on" prefix (same for "onanimationiteration")

nghuuphuoc: My bad! I just updated the PR.

raquo: No problem @nghuuphuoc! Thanks for taking the time to add these here.

Laminar v0.12.0 will be out in a few days. I will publish your changes to _Scala DOM Types_ at that time too (it's a breaking change due to the added type param, so even if I published sooner, you wouldn't be able to use that from Laminar until a new Laminar is released).

PR #61, opened 2020-12-23
Added support for dom.TransitionEvent

ajaychandran:
The [transitionrun](https://developer.mozilla.org/en-US/search?q=transitionrun) and [transitionstart](https://developer.mozilla.org/en-US/search?q=transitionstart) events were not added since they are not defined in the [GlobalEventHandlers](https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers) mixin.


raquo:
Thanks for the PR!

I think if we're adding transition events, we should add all of them. Being part of GlobalEventHandlers is not a requirement for inclusion in MiscellaneousEventProps.scala, it's basically a grab bag of random events that didn't fit elsewhere and that didn't justify creating a separate trait for.

Also, MDN  says that Chrome doesn't support transition events, only Firefox and Safari do. Strictly speaking a property doesn't have to be fully supported by all browsers to be included in SDT, but is it actually useful in real life, when the most popular browser doesn't support it? Do you know if maybe the Chrome team is planning to add support for this soon?

ajaychandran:
Actually, only the `transitioncancel` event is not available on Chrome. The same is true for [animationcancel](https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/onanimationcancel) event that was added in #60 .

The `transitionend` event is more widely [supported](https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/ontransitionend#Browser_compatibility). 
The property name differs across browsers (`ontransiitonend` vs `onwebkittransitionend`) but the event name, `transitionend`, is consistent (this is what is actually used for listener registration).

I verified this on the Chrome (Version 87.0.4280.88) homepage using **Developer Tools > Console**.
```js
node = document.getElementById("realbox");
node.style.transition = "opacity 2s";
node.addEventListener('transitionend', function(e) { console.log(e); })
node.style.opacity = "0.2";
// after 2s
TransitionEvent {isTrusted: true, propertyName: "opacity", elapsedTime: 2, pseudoElement: "", type: "transitionend", …}
```
I propose to retain only the `transitionend` event. The other transition events (and `animationcancel` event) should be excluded due to lack of support across browsers.

WDYT?

ajaychandran:
Added `experimental` object in `MiscellaneousEventProps` with the following events:

- `onAnimationCancel`
- `onTransitionCancel`
- `onTransitionRun`
- `onTransitionStart`

This way the features are available and their limited support is made explicit.

Please review.

raquo:
Hm, I don't think having an `experimental` object for these is the way to go. Because of how these traits are mixed into a singleton object, this would prevent having similar `experimental` objects in other traits. Also, I just don't see many people building web apps that don't work in Chrome, and those few who do, can define these props in their own projects.

I think we're better off not including those four events in Scala DOM Types at all.

@nghuuphuoc We're gonna remove `onAminationCancel` event that you added because MDN says it's not supported by Chrome (if it's out of date, please let me know and we'll keep it). It's the most popular browser so I doubt many people will want to use it. On the flip side, more people will be confused why it doesn't work if we do include it.

ajaychandran:
Note that `onAnimationCancel` has been removed. 
The `animationcancel` event is [working](https://www.w3schools.com/code/tryit.asp?filename=GM153OSD2CV6) in Chrome (Version 87.0.4280.88) but I think its safer to exclude it at this time.

raquo:
Thanks, just checked Chromium 81 with your code, `animationcancel` event doesn't work there. Looks like support was added [very recently](https://www.chromium.org/teams/animations/highlights-archive/march-2020---web-animations-scroll-linked-animations-throughput-metrics-and-more). I'm fine omitting that event for now. It's easy to add if anyone asks for it. I'll merge this PR soon.

PR #64, opened 2021-01-16
Initial attempt to upgrade to scala3 rc1

yurique:
based on top of scalavisions changes: https://github.com/scalavision/scala-js-dom/compare/upgrade-scala3-M3...yurique:upgrade-scala3-M3?expand=1


build.sbt
```diff
@@ -1,96 +1,36 @@
...
-  .jsSettings(
-    scalaJSLinkerConfig ~= { _.withSourceMap(false) },
-    requireJsDomEnv in Test := true,
-    useYarn := true,
-    libraryDependencies ++= Seq(
-      "org.scala-js" %%% "scalajs-dom" % "1.0.0"
+lazy val root =
+  project.in(file("."))
+    .aggregate(domtypes.js, domtypes.jvm)
+    .settings(noPublish)
```
raquo, line 33: I guess it's not possible to configure noPublish from release.sbt? Not sure which way is better, it just seems everything else release-related is there now.

build.sbt
```diff
@@ -1,96 +1,36 @@
...
+
+lazy val domtypes =
+  crossProject(JSPlatform, JVMPlatform).in(file("."))
+    .settings(
+      scalacOptions ~= (_.filterNot(Set(
+        "-Wunused:params",
+        "-Ywarn-unused:params",
+        "-Wunused:explicits"
+      ))),
+      scalacOptions in (Compile, doc) ~= (_.filter(_.startsWith("-Xplugin"))),
```
raquo, line 12: What / why are you filtering for here?

build.sbt
```diff
@@ -1,96 +1,36 @@
-import sbtcrossproject.CrossPlugin.autoImport.crossProject
-
-inThisBuild(Seq(
-  name := "Scala DOM Types",
-  normalizedName := "domtypes",
```
raquo, line 5: Where does the new config take `normalizedName` from? I'd prefer for it to be explicit.

build.sbt
```diff
@@ -1,96 +1,36 @@
...
-    requireJsDomEnv in Test := true,
-    useYarn := true,
-    libraryDependencies ++= Seq(
-      "org.scala-js" %%% "scalajs-dom" % "1.0.0"
+lazy val root =
+  project.in(file("."))
+    .aggregate(domtypes.js, domtypes.jvm)
+    .settings(noPublish)
+    .settings(
+      name := "Scala DOM Types"
```
raquo, line 35: This used to be in `ThisBuild / name`, right? I'm sbt-challenged, is this the same, or if not, what's the difference?

build.sbt
yurique: yeah, can't refer a `lazy val` defined in a different sbt file

build.sbt
yurique: it's because of this: https://github.com/DavidGregory084/sbt-tpolecat/issues/36

I later changed this to explicitly filtering out the offensive scalac options, but forgot to change here: https://github.com/raquo/Airstream/pull/68/files#diff-5634c415cd8c8504fdb973a3ed092300b43c4b8fc1e184f7249eb29a55511f91R24-R39

build.sbt
yurique: sbt says it's "not used by any other settings/tasks", so it doesn't affect anything

build.sbt
yurique: With a single module (`root`) the result is the same (if there were modules, `ThisBuils / name` would set the same name for all of them). In this case, I don't really have a strong opinion (just did it as it is in Airstream/Laminar).

PR #75, opened 2021-12-24
API: Complete rework of CSS Styles; Also nuke everything else that is not strictly required

### LINKED ISSUE #2: Extract StyleSetters (eg. for colors) into traits and mix them into revelant objects
The StyleSetters for the css attribute `color` are also relevant for other css attributes accepting the [color property](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value) like `background-color`, `border-color`, `box-shadow`, `outline-color`, `text-shadow`.

I think code sharing makes sense here.

https://github.com/raquo/scala-dom-types/blob/06dab597ead3eb9eba267efeff10eacb49fe352b/shared/src/main/scala/com/raquo/domtypes/generic/defs/styles/Styles.scala#L505

(I'm writing down ideas I have while porting a project to the current outwatch PR.)

raquo:
Fixes #2

It has been a long standing issue that SDT creates instances of `Style` props internally. The goal of SDT is to provide type listings in the most flexible manner, it should not require usage of any concrete types for keys. To that effect I refactored how we deal with styles to be similar to how we deal with attrs and props, with adaptations needed to provide well typed listings of CSS keywords that enable syntax like `display.none` (instead of `display := "none")` in the consuming libraries. I've also sorted the props in a sane manner, fixed a couple typos, added links to MDN docs and otherwise cleaned stuff up.

Not only that, I also added a concept of derived style props, enabling APIs like `marginTop.px := 12` instead of `marginTop := "12px"`. As it stands right now, this is a non-optional part of the new SDT types – an implementation to support that is **required** in the consuming library. If anyone cares to object, I can probably make it optional at the cost of a bit of boilerplate in the consuming library's setup.

SDT also contained a bunch of common value types that I thought the consuming libraries might want to share, such as Attr, Element, Modifier, etc. In a few years that this library existed, that didn't happen, and I don't think it will, so I moved all those types into test fixtures. If you were using some of those types, you'll need to implement them in your library (it's trivial, copy-paste away).

# TODO
- [x] Show Laminar and scala-dom-testutils migration for this change (soon)
- [x] Get a nod from Outwatch that this is fine
- [x] Document the changes
- [x] Clean up scaladoc links to old `keys.`, `nodes.`, `canonical.`, and other moved files.
- [x] Check new CSS props for typos and copy-paste errors


raquo:
Scala DOM TestUtils diff: https://github.com/raquo/scala-dom-testutils/pull/6/commits/68926e1fe76a696429d895d679a2a37012f8d2c2

Laminar diff: https://github.com/raquo/Laminar/commit/0329c3d75dfe66489ce01aa669e1b8730f13581c

raquo:
@fdietze @cornerman Hey guys, please check this change out for Outwatch purposes. You can see the crux of the changes when it comes to library integration in CompileTest, where it says "with Styles", and I also posted Laminar's adaptation diff for this change in the previous comment here.

I would like to merge this in a couple weeks if there are no objections.

fdietze:
Amazing work @raquo, thank you very much! :+1: 
I hope I find the time to test it in outwatch soon.

fdietze:
So I tried porting outwatch to this PR, but while doing this, I realized that I don't have enough knowledge about the integration to fix things quickly. I'm sorry, I guess I have to pass this on to @cornerman ...

cornerman:
Hi all! I am on it and have most things done already locally - just needs some finishing touches. Overall, this will work well with outwatch. My only concern is whether we could somehow reduce the boilerplate of instantiating all `new ... with keywords...` for the different styles. Will give an update again when I am finished with the changes!

Thank you so much @raquo for taking care of this issue :)

raquo:
Cool!

> My only concern is whether we could somehow reduce the boilerplate of
instantiating all new ... with keywords... for the different styles.

Sadly, I'm afraid that we as the library authors have to deal with it. This
boilerplate was the main reason why I didn't implement it like this
initially – unfortunately Scala just doesn't seem to have a better way of
abstracting over class declarations. Well, to be fair, we could probably
achieve a similar-ish public API with implicit extension methods, I guess,
but that would be too inscrutable for my taste. For a low level library
dumber is better imo.

raquo:
I will release this as 0.16.0-RC1 in a moment, if you face any problems with it, please let me know.

raquo:
Released 0.16.0-RC2 with minor fixes. This will become 0.16.0 if no issues are found in the next couple weeks or so.

PR #76, opened 2022-01-12
Add download attribute

ioleo:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#attr-download


raquo:
Nice, thanks! Sorry for the delay, big Laminar changes afoot. This will go into SDT 0.16.0, and will be available in Laminar 0.15.0 when it's released "in a few weeks".

PR #77, opened 2022-02-07
Add aria-current attribute.

mprihoda:
Adding aria-current to AriaAttrs.


raquo:
Thanks! Sorry, took me a while.

PR #78, opened 2022-03-21
Remove unused type parameter from `ComplexHtmlKeys`

armanbilge:
Thanks for a really fantastic lib btw :)


raquo:
Thanks! Yes, looks like I forgot to remove the type param when removing the `text` property from this trait.

Since this bug doesn't actually break anything, I will merge this PR with the next major version.

PR #80, opened 2022-05-13
Add svg title tag

Lasering:


raquo:
Thanks! Hm, I think maybe we should call this `titleTag` instead of `title`, to match the name that we use for HTML title tag? (the name `title` is taken by the title attribute in the html namespace)

Lasering:
Done

raquo:
Thanks!

PR #83, opened 2022-08-31
Integrate with `InputEvent`

armanbilge:
Finally added this 😅

I'm a bit confused, it wasn't obvious to me if `DomInputEvent` is being used anywhere?
https://github.com/raquo/scala-dom-types/blob/de166bd3816ae2ad35bbc2fb405a120e46a03afb/shared/src/main/scala/com/raquo/domtypes/generic/defs/eventProps/FormEventProps.scala#L24-L33

### DIFF
--- a/shared/src/main/scala/com/raquo/domtypes/generic/defs/eventProps/FormEventProps.scala
+++ b/shared/src/main/scala/com/raquo/domtypes/generic/defs/eventProps/FormEventProps.scala
@@ -20,7 +20,6 @@ import com.raquo.domtypes.generic.builders.EventPropBuilder
   *            - Similarly for `onChange` and `onSelect` – these could also fire on an `HTMLTextAreaElement`.
   * @tparam DomInputEvent
   *            DOM InputEvent https://developer.mozilla.org/en-US/docs/Web/API/InputEvent
-  *            Note: This type is not implemented in scala-js-dom
   */
 trait FormEventProps[
   EP[_ <: DomEvent],
--- a/shared/src/main/scala/com/raquo/domtypes/generic/defs/eventProps/FormEventProps.scala
+++ b/shared/src/main/scala/com/raquo/domtypes/generic/defs/eventProps/FormEventProps.scala
@@ -49,6 +49,17 @@ trait FormEventProps[
   lazy val onSelect: EP[DomHtmlElementTargetEvent] = eventProp("select")
 
   /**
+    * The DOM beforeinput event fires when the value of an <input>, or <textarea>
+    * element is about to be modified. The event also applies to elements with
+    * contenteditable enabled, and to any element when designMode is turned on.
+    *
+    * MDN
+    *
+    * @note IE does not support this event.
+    */
+  lazy val onBeforeInput: EP[DomInputEvent] = eventProp("beforeinput")
+
+  /**
     * The input event is fired for input, select, textarea, and
     * contentEditable elements when it gets user input.
     */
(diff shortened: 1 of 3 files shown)


raquo:
Nice, thanks!

> I'm a bit confused, it wasn't obvious to me if DomInputEvent is being used anywhere?

Huh, I figured that would be for the `input` event, but no, apparently [it's just a generic Event](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/input_event).

The only event I can find that uses InputEvent is [beforeinput](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/beforeinput_event), but currently we don't have it defined because it's rarely used and not supported by IE.

But well, since we're here, might as well add `onBeforeInput` and use the type param? If you do, please add `Note: IE does not support this event` on a new line in its ScalaDoc.

armanbilge:
I think the problem is that IE just doesn't support `InputEvent`.
https://caniuse.com/mdn-api_inputevent

For example the note on this page suggests that an `InputEvent` is indeed used for `input`.

> On browsers that don't fully implement the [InputEvent](https://developer.mozilla.org/en-US/docs/Web/API/InputEvent) interface which is used for the [beforeinput](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/beforeinput_event) and [input](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/input_event) events, you may get incorrect output on those lines of the log output.

https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key#result

Ah, here is some explanation.

> For [<textarea>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/textarea) and [<input>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input) elements that accept text input (type=text, type=tel, etc.), the interface is [InputEvent](https://developer.mozilla.org/en-US/docs/Web/API/InputEvent); for others, the interface is [Event](https://developer.mozilla.org/en-US/docs/Web/API/Event).

https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/input_event

Ok well I don't think there's anything we can do to make that more typesafe is there?

I will add `onBeforeInput` as you suggested, thanks!

shared/src/main/scala/com/raquo/domtypes/generic/defs/eventProps/FormEventProps.scala
```diff
@@ -49,6 +48,17 @@ trait FormEventProps[
...
+  /**
+    * The DOM beforeinput event fires when the value of an <input>, or <textarea>
+    * element is about to be modified. The event also applies to elements with
+    * contenteditable enabled, and to any element when designMode is turned on.
+    *
+    * MDN
+    *
+    * @note IE does not support this event.
+    */
+  lazy val onBeforeInput: EP[DomInputEvent] = eventProp("beforeinput")
```
armanbilge, line 60: I wasn't sure how you like to style these mixed case here. This seemed sensible.

raquo:
> For [<textarea>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/textarea) and elements that accept text input (type=text, type=tel, etc.), the interface is [InputEvent](https://developer.mozilla.org/en-US/docs/Web/API/InputEvent); for others, the interface is [Event](https://developer.mozilla.org/en-US/docs/Web/API/Event).

Sigh... Don't you just love the JS DOM API? Thanks for looking into it.

PR looks good to me, I will merge and release it when I get back to OSS work in October. (I try to keep `master` free of unreleased changes even if they're approved)

Issue #87, opened 2022-11-19
RFC: Use concrete types instead of generic type params?


raquo:
Currently, the design of SDT type listings is maximally generic, providing a lot of flexibility to esoterically minded library authors.

The issues below mostly apply to all types of keys – props, attrs, css style props, and event props, but we'll be looking at specific event props as a concrete example:

```scala
trait MouseEventProps[
  EP[_ <: DomEvent],
  DomEvent,
  DomMouseEvent <: DomEvent,
  DomElementMouseEvent <: DomMouseEvent,
  DomDragEvent <: DomMouseEvent,
  DomWheelEvent <: DomMouseEvent
] { this: EventPropBuilder[EP, DomEvent] =>

    /**
    * The click event is raised when the user clicks on an element. The click
    * event will occur after the mousedown and mouseup events.
    *
    * MDN
    */
  lazy val onClick: EP[DomElementMouseEvent] = eventProp("click")

  // ... more props here

  /**
    * Fires when the mouse wheel rolls up or down over an element
    */
  lazy val onWheel: EP[DomWheelEvent] = eventProp("wheel")

  // ... more props here
}
```

A developer who uses a UI library like Laminar (or Outwatch or Calico) might see `onWheel --> whatever` in their code, and ctrl-click on the `onWheel` symbol in their IDE to go to its definition. The snippet above is where they will land.

The definition of `onWheel` has a short ScalaDoc comment, but there is no concrete information about the value we're dealing with. We don't know what this `EP` type is, or what shape `DomWheelEvent` is – these types are both abstract.

---

We could improve on `DomWheelEvent` by replacing it with `org.scalajs.dom.WheelEvent` – the concrete type that it resolves to for 99% of the users. Then users could proceed to ctrl-click into that type and see what kind of events they will get. This would improve ergonomics for most people. This benefit is obvious, the main issue is maintaining functionality for the other 1%, the use cases where we do want abstract type params.

---

Currently `DomWheelEvent` is an abstract type param because our `trait MouseEventProps` is in a `shared` project – it's cross-compiled for Scala.js and the JVM, and so it does not have access to the JS-only `org.scalajs.dom` types.

I designed it that way  because I was hoping that Scala DOM Types would be used on the JVM as well, for example in **tools that need to render HTML (web pages or email content) on the server**. To my knowledge, as of today no such tools exist that use _Scala DOM Types_, but it's possible that they do exist, just that their authors are overly modest, and didn't let us know.

[...]

PR #89, opened 2022-11-28
Refactor everything in SDT for compile-time code generation

raquo:
See #87 for rationale and discussion.

This PR is taking the second approach – consuming libraries will call an SDT method at compile time to generate the code of all the traits that previously lived in SDT (`HtmlAttrs`, `Props`, `Styles`, etc.) They can also write their own code generators.

In the initial commit, I made parser to parse existing SDT traits and generate the code of the new SDT. The generated code is in `src/main/scala/com/raquo/domtypes/shared/defs`, and the new type definitions are in `src/main/scala/com/raquo/domtypes/shared/common`. Basically everything else in this project will be nuked.

Lots of stuff is still WIP, but if you're interested you might want to look in the two directories mentioned above.

Next step for me is writing the code generators that will take this new SDT data and generate the source code of the traits that you would want to actually use in third party libraries.

SDT also used to contain a few small helpers like `Codec`-s, `SvgNamespaces,` and `VendorPrefixes`. I'm not yet sure if I will keep those in SDT, or if each library will just need to copy-paste a few files. These particular helpers don't really ever change.


raquo:
I've implemented most generators that will generate Laminar / Outwatch / etc. source code from the new SDT data. For some examples:

1) SDT source data now looks like this: [Props](https://github.com/raquo/scala-dom-types/blob/codegen/shared/src/main/scala/com/raquo/domtypes/defs/props/Props.scala) and [ReflectedAttrs](https://github.com/raquo/scala-dom-types/blob/codegen/shared/src/main/scala/com/raquo/domtypes/defs/reflectedAttrs/ReflectedHtmlAttrs.scala)

2) In your library, you call the relevant generator at compile time to produce a trait like this: [Props](https://github.com/raquo/scala-dom-types/blob/codegen/js/src/test/scala/com/thirdparty/defs/Props.scala) (this combines props and reflected-attrs-implemented-as-props in one trait)

3) In your library, you provide a few of your own types like [Attr](https://github.com/raquo/scala-dom-types/blob/codegen/js/src/test/scala/com/thirdparty/keys/Attr.scala), [Codec](https://github.com/raquo/scala-dom-types/tree/codegen/js/src/test/scala/com/thirdparty/codecs), etc.

4) When all is said and done, you end up with a file structure like [this example](https://github.com/raquo/scala-dom-types/tree/codegen/js/src/test/scala/com/thirdparty) in your own project. It should be very similar to what you already have.

The code generation is configurable / extendable, but not infinitely so. At some point it might be easier to write your own non-generic generator, but I expect all reasonable frontend use cases to be covered by the canonical generator hosted in SDT. If some required customization is missing, it can be added later.

The traits in [thirdparty/defs](https://github.com/raquo/scala-dom-types/tree/codegen/js/src/test/scala/com/thirdparty/defs) are currently generated by [GeneratorSpec](https://github.com/raquo/scala-dom-types/blob/codegen/jvm/src/test/scala/com/raquo/domtypes/GeneratorSpec.scala) as an example.

I still need to finish the parsing and generation of CSS styles, and work on edge cases like composite keys, aliases, etc.

raquo:
FYI I'm also dropping Scala 2.11 support, seems like a waste of cpu cycles nowadays.

raquo:
Dropped built-in support for TypedTargetEvent. I think it only complicated type signatures, and for little gain, since the best type we could infer would be `dom.html.Element` (not e.g. dom.html.Input, which would let you access `.value`). It can't be made both type safe and more useful even if it's complicated even further, so down the drain it goes.

If you really want to support `TypedTargetEvent` in your own library, the new generators should be flexible enough to allow that. For myself, I plan to ditch TypedTargetEvent in Laminar in favor of the type-safe `inContext { thisNode =>` and unsafe `mapToTargetAs[elementType]`.

--

Also, temporarily dropped support for the `tupled` convenience method on the `display` style prop: `display.tupled("inline", "flex")`. It was half-baked, and I will eventually develop a more useful and more generic version, but not right now.

PR #90, opened 2023-01-12
Backport "New: download reflected attribute (#76)"

lolgab:
I'm still on laminar 0.14.x and using the download attribute in anchors.
For now, I'm doing:
```scala
link.ref.asInstanceOf[js.Dynamic].download = "result.zip"
```
but it would be nice if the addition could be backported to `domtypes` `0.15`


raquo:
Sorry, I was a bit abrupt. I appreciate the effort to backport of course. It's just, the time it will take me to do all the release steps will be more than the total time it will take _all_ developers who need this attribute in 0.15.x to add that one line to their code. IMO software is all about making use of leverage (1x effort in – 10x result out), and there's just no leverage here.

lolgab:
Makes sense! Thank you! I did know the workaround but had forgot it :) Yeah, with `customProp` it's much cleaner!

PR #93, opened 2023-01-14
Support modifiers for generated `trait`s

armanbilge:
e.g. `private[package]`, `sealed`, etc.


shared/src/main/scala/com/raquo/domtypes/codegen/generators/PropsTraitGenerator.scala
```diff
@@ -10,6 +10,8 @@ class PropsTraitGenerator(
   override protected val headerLines: List[String],
   override protected val traitCommentLines: List[String],
   override protected val traitName: String,
+  override protected val traitModifiers: List[String],
+
```
armanbilge, line 14:
```suggestion
  override protected val traitModifiers: List[String],
```

shared/src/main/scala/com/raquo/domtypes/codegen/generators/EventPropsTraitGenerator.scala
```diff
@@ -8,6 +8,8 @@ class EventPropsTraitGenerator(
   override protected val defGroupComments: EventPropDef => List[String],
   override protected val headerLines: List[String],
   override protected val traitCommentLines: List[String],
+  override protected val traitModifiers: List[String],
+
```
armanbilge, line 12:
```suggestion
  override protected val traitModifiers: List[String],
```

armanbilge:
CI is a bit flaky, seems like there is some sort of race condition.

raquo:
Thanks!

CompileSpec requires the files generated by GeneratorSpec. The order of platforms is set correctly when running tests from the root project, but I think the problem is that `writeToFile` does not wait for the file to be written to disk before returning (apparently flushing the stream is not enough?). I've addressed that, let's hope this fixes it.

PR #94, opened 2023-01-20
Escape html tags in scaladoc

armanbilge:
They are breaking Scaladoc 3, at least.

For example. scroll to the bottom of this page to see the scaldoc for `cols` (that's where it ends, because that's where it breaks 😅 )
https://s01.oss.sonatype.org/service/local/repositories/snapshots/archive/com/armanbilge/calico-docs_sjs1_3/0.2-100451d-SNAPSHOT/calico-docs_sjs1_3-0.2-100451d-SNAPSHOT-javadoc.jar/!/calico/html/Html.html

### DIFF
--- a/js/src/test/scala/com/thirdparty/defs/attrs/HtmlAttrs.scala
+++ b/js/src/test/scala/com/thirdparty/defs/attrs/HtmlAttrs.scala
@@ -82,7 +82,7 @@ trait HtmlAttrs {
 
 
   /**
-    * The max attribute specifies the maximum value for an <input> element of type
+    * The max attribute specifies the maximum value for an `<input>` element of type
     * number, range, date, datetime, datetime-local, month, time, or week.
     * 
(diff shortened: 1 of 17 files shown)


raquo:
Hm... Scaladoc renderers parse raw HTML? TIL.

And I guess scaladoc generation does not fail because this step does not validate the HTML in every comment block, so we can't really detect this at compile time (elegantly, at least).

raquo:
Thanks, released as 17.0.0-M5

armanbilge:
Thanks, seems to have fixed it!
https://s01.oss.sonatype.org/service/local/repositories/snapshots/archive/com/armanbilge/calico-docs_sjs1_3/0.2-7e28afb-SNAPSHOT/calico-docs_sjs1_3-0.2-7e28afb-SNAPSHOT-javadoc.jar/!/calico/html/Html.html

PR #100, opened 2023-11-06
Feature: add crossorigin attr in ReflectedHtmlAttrDef

cyz1901:
Add crossorigin attr in ReflectedHtmlAttrDef
https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/crossorigin

### DIFF
--- a/shared/src/main/scala/com/raquo/domtypes/defs/reflectedAttrs/ReflectedHtmlAttrDefs.scala
+++ b/shared/src/main/scala/com/raquo/domtypes/defs/reflectedAttrs/ReflectedHtmlAttrDefs.scala
@@ -1106,9 +1106,9 @@ object ReflectedHtmlAttrDefs {
     ),
 
     ReflectedHtmlAttrDef(
-      scalaName = "crossorigin",
+      scalaName = "crossOrigin",
       domAttrName = "crossorigin",
-      domPropName = "crossorigin",
+      domPropName = "crossOrigin",
       scalaValueType = "String",
       domPropValueType = "String",
       attrCodec = "StringAsIs",
@@ -1119,7 +1119,9 @@ object ReflectedHtmlAttrDefs {
         "be loaded when requested from a different domain, and how to handle potential",
         "security issues, like CORS (Cross-Origin Resource Sharing) policies.",
         "The value of this attribute determines whether the browser will allow or",
-        "block loading of the resource, helping to enhance web security."
+        "block loading of the resource, helping to enhance web security.",
+        "",
+        """Allowed values: "anonymous" | "use-credentials" | "" (same as "anonymous") """
       ),
       docUrls = List(
         "https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/crossorigin"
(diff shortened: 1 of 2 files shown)


shared/src/main/scala/com/raquo/domtypes/defs/reflectedAttrs/ReflectedHtmlAttrDefs.scala
```diff
@@ -1105,6 +1105,27 @@ object ReflectedHtmlAttrDefs {
       docUrls = Nil,
     ),
 
+    ReflectedHtmlAttrDef(
+      scalaName = "crossorigin",
```
raquo, line 1109: This should be camelCased to follow our naming conventions: "crossOrigin"

shared/src/main/scala/com/raquo/domtypes/defs/reflectedAttrs/ReflectedHtmlAttrDefs.scala
```diff
@@ -1105,6 +1105,27 @@ object ReflectedHtmlAttrDefs {
...
+      domAttrName = "crossorigin",
+      domPropName = "crossorigin",
```
raquo, line 1111: `domPropName` should be camelCased to "crossOrigin" because that's what it's called in JS DOM (it won't work otherwise). I know, it's a bit weird – JS reflected properties can have names that are slightly different from the corresponding attribute names.

shared/src/main/scala/com/raquo/domtypes/defs/reflectedAttrs/ReflectedHtmlAttrDefs.scala
```diff
@@ -1105,6 +1105,27 @@ object ReflectedHtmlAttrDefs {
...
+      domPropValueType = "String",
+      attrCodec = "StringAsIs",
+      propCodec = "StringAsIs",
+      commentLines = List(
+        "This attribute specifies how the browser should handle cross-origin",
+        "requests for the associated resource. It controls whether the resource can",
+        "be loaded when requested from a different domain, and how to handle potential",
+        "security issues, like CORS (Cross-Origin Resource Sharing) policies.",
+        "The value of this attribute determines whether the browser will allow or",
+        "block loading of the resource, helping to enhance web security."
```
raquo, line 1122: Please add two more lines to the docs:
"",
"""Allowed values: "anonymous" | "use-credentials" | "" (same as "anonymous") """

raquo (changes requested):
Thanks! Can you please also run the tests locally before commiting – this will update some generated files that we keep in git.

shared/src/main/scala/com/raquo/domtypes/defs/reflectedAttrs/ReflectedHtmlAttrDefs.scala
cyz1901: fixed

shared/src/main/scala/com/raquo/domtypes/defs/reflectedAttrs/ReflectedHtmlAttrDefs.scala
cyz1901: fixed

shared/src/main/scala/com/raquo/domtypes/defs/reflectedAttrs/ReflectedHtmlAttrDefs.scala
cyz1901: fixed

PR #101, opened 2023-11-14
Add anywhere value to overflow-wrap style

FabioPinheiro:
https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap


raquo:
Thanks!

PR #105, opened 2024-09-20
WIP: Math ml tag defs

Quafadas:
I will not pretend to have written any of this myself - thank you Mr AI.

In the light of the above statement, I'm not clear whether I have a better or worse chance of this actually working. 

So let's poke the CI and see what happens.


davidon-top:
as far as i know org.scalajs.dom doesn't have support for mathml so your `scalaJsElementTypeAlias` assignments wont work after codegen, i can be mistaken tho i'm not certain

Quafadas:
You beat the CI to it! 

https://github.com/Quafadas/scala-dom-types/actions/runs/10959385719/job/30431627724

I'm actually not sure whether it needs to be supported by `org.scalajs.dom` explicitly though. 

The browser seems to render it quite happily, so I'm wondering if I just go for some base class and pray?

Quafadas:
I guess this is probably going to impact the attributes it can have though? Does that mean I need to make PR to scala-js-dom?

davidon-top:
by default sdt generates definitions like these `lazy val htmlRootTag: HtmlTag[dom.HTMLHtmlElement] = htmlTag("html")` and u can see inside the HtmlTag generic the dom.whatever that is what `scalaJsElementTypeAlias` describes AFAIK, its probably possible to add it without scala-js-dom having mathml tags but it will need to use a different system/design to make the mathml elements

davidon-top:
> I guess this is probably going to impact the attributes it can have though? Does that mean I need to make PR to scala-js-dom?

i don't think it would impact the attributes since they are basically just a name and codec it would impact tags tho since they accept the generic type from org.scalajs.dom

Quafadas:
So I've pattern matched my way through the CI, which is green on my fork. Hopefully it's a starting point for a discussion.

Quafadas:
I assume, that in its curren t form, it will exhibit the same set of problems as the original bootlegged HTML. I'm not clear how to work around that though.

jvm/src/test/scala/com/raquo/domtypes/GeneratorSpec.scala
```diff
@@ -90,6 +90,31 @@ class GeneratorSpec extends AnyFunSpec with Matchers {
...
+    val fileContent = generator.generateTagsTrait(
+      tagType = MathMlTagType,
+      defGroups = defGroups.mathMlTagsDefGroups,
+      printDefGroupComments = false,
+      traitCommentLines = Nil,
+      traitModifiers = Nil,
+      traitName = traitName,
+      keyKind = "MathTag",
+      baseImplDefComments = Nil,
+      keyImplName = "mathTag",
```
raquo: `mathTag` name is already taken for the <math> tag, this should be something else. I think`keyImplName` should be `mathMlTag`.

shared/src/main/scala/com/raquo/domtypes/codegen/CanonicalDefGroups.scala
```diff
@@ -25,6 +25,10 @@ class CanonicalDefGroups {
     "SVG Tags" -> SvgTagDefs.defs,
   )
 
+  val mathMlTagsDefGroups: List[(String, List[TagDef])] = List(
+    "MathMl Tags" -> MathMlTagDefs.defs,
```
raquo, line 29: "MathMl" should be capitalized as "MathML" in the key here, to match other keys that capitalize e.g. SVG.

shared/src/main/scala/com/raquo/domtypes/codegen/CanonicalGenerator.scala
```diff
@@ -168,10 +168,16 @@ class CanonicalGenerator(
...
+    } else  if (tagType == SvgTagType) {
       List(
         s"def ${keyImplName}[$scalaJsElementTypeParam <: $baseScalaJsSvgElementType]($keyImplNameArgName: String): ${keyKind}[$scalaJsElementTypeParam] = ${keyKindConstructor(keyKind)}($keyImplNameArgName)",
       )
+    } else if (tagType == MathMlTagType) {
+      List(
+        s"def ${keyImplName}[$scalaJsElementTypeParam <: $baseScalaJsHtmlElementType]($keyImplNameArgName: String): ${keyKind}[$scalaJsElementTypeParam] = ${keyKindConstructor(keyKind)}($keyImplNameArgName)",
+      )
+    } else {
+      ???
```
raquo, line 182: Yeah that makes more sense, it was too loose.

raquo (commented):
Thanks, that's a good start!

After looking into this a bit, I think overall, we need to treat MathML tags and attributes similarly to how we treat SVG – so they will get their own namespace in Laminar.

Indeed, as you said, for this to be meaningful, someone needs to contribute the [MathMLElement](https://developer.mozilla.org/en-US/docs/Web/API/MathMLElement) interface to scalajs-dom. I think we can do without that if we just use the `dom.Element` supertype for MathML elements, but that's only suitable for a PoC, it would make the types too loose in Laminar.

I think if we're adding MathML, we should add all of the MathML tags and attributes right away, especially if AI is helping with the chore. Also:
* The tags and attributes should be sorted alphabetically by their JS / DOM name in the listings
* The `<math>` tag itself should be moved to MathMlTags – it's actually a MathMlElement, not an HTMLElement (similar to the <svg> root tag)

I made an issue in Laminar to track all the related tasks: https://github.com/raquo/Laminar/issues/172

Quafadas:
@raquo This passes on my fork, I'm not in a position to judge, if the tests are meaningful or not, however :-(

.vscode/settings.json
```diff
@@ -0,0 +1,5 @@
+{
+  "files.watcherExclude": {
+    "**/target": true
+  }
+}
```
raquo: I'm not familiar with VS Code – did VS Code create this file on its own, or did you need to specify this exclusion yourself? If VS code creates such files automatically without user input, then we should probably gitignore the .vscode dir.

build.sbt
```diff
@@ -12,9 +14,11 @@ ThisBuild / dynver := {
   sbtdynver.DynVer
     .getGitDescribeOutput(d)
     .mkVersion(out => versionFmt(out, dynverSonatypeSnapshots.value), fallbackVersion(d))
+
+
 }
 
-(ThisBuild / scalaVersion) := Versions.Scala_2_12
+(ThisBuild / scalaVersion) := Versions.Scala_3
```
raquo: This should stay as 2.12 because SDT is typically used from sbt which runs 2.12.

I linked to Scala 3 scalajs-dom snapshots because that was the exact link that I found, but the 2.12 snapsohts are also published, just need to change the url to https://oss.sonatype.org/content/repositories/snapshots/org/scala-js/scalajs-dom_sjs1_2.12/

js/src/test/scala/com/thirdparty/tags/Tag.scala
```diff
@@ -14,3 +15,7 @@ case class HtmlTag[El <: dom.html.Element](
 case class SvgTag[El <: dom.svg.Element](
   override val domName: String
 ) extends Tag[El]
+
+case class MathTag[El <: MathMLElement](
```
raquo: Shoud be called `MathMlTag`

js/src/test/scala/com/thirdparty/tags/Tag.scala
```diff
@@ -14,3 +15,7 @@ case class HtmlTag[El <: dom.html.Element](
...
+  override val domName: String
+) extends Tag[org.scalajs.dom.MathMLElement]
```
raquo: This doesn't need `org.scalajs.dom.` prefix, you already imported it

jvm/src/test/scala/com/raquo/domtypes/GeneratorSpec.scala
```diff
@@ -90,6 +90,31 @@ class GeneratorSpec extends AnyFunSpec with Matchers {
     )
   }
 
+  it("Generate MathMl tags") {
+    println("=== MATHML TAGS ===")
+
+    val traitName = "MathMLTags"
```
raquo: "ML" should be TitleCased as "Ml" for all the Scala DOM Types names, similarly to how we treat HTML and SVG abbreviations. Please change in other names too like mathMLTag

shared/src/main/scala/com/raquo/domtypes/defs/tags/MathMlTagDefs.scala
```diff
@@ -0,0 +1,334 @@
...
+import com.raquo.domtypes.common.{MathMLTagType, TagDef}
+
+
+object MathMLTagDefs {
+
+    val defs: List[TagDef] = List(
+
+      TagDef(
+        tagType = MathMLTagType,
+        scalaName = "math",
```
raquo: `math` is already a built-in Scala package, this should be called `mathTag` in Scala.

shared/src/main/scala/com/raquo/domtypes/defs/attrs/MathMLDefs.scala
```diff
@@ -0,0 +1,22 @@
...
+    * This would appear to be because they aren't part of the browser API?
+    *
+    * We take a pass on these here.
+    *
+    */
+val defs: List[AttrDef] = List(
+
+
+  )
+}
```
raquo, line 22: No – there's a difference between markup attributes and DOM properties. The properties are what's missing – as evidenced by `undefined` in the browser, as we've seen. But the attributes of course still exist. I'm not proficient in mathml, but I assume that specifying them is in fact needed to accomplish some things.

Markup attributes are typically specified in the mathml markup, but they can also be set via JS, like this: `$0.setAttribute('displaystyle', 'normal')` (You can try that on [this page](https://developer.mozilla.org/en-US/docs/Web/MathML/Element/mstyle), if you inspect & select the `<mstyle>` element in their example – you will see the formula change its layout.

Looks like this is the list of attributes: https://developer.mozilla.org/en-US/docs/Web/MathML/Attribute

Don't add the ones with the trash icon.

With the attributes, the most important thing is to get their types /codecs right. See the types in the table above.
- If the type is not mentioned, it's just a string, so for them the codec would be just StringAsIs.
- At the moment we encode enum-type values as strings as well – just specify their allowed values in the scaladoc.
- For booleans, [this page](https://developer.mozilla.org/en-US/docs/Web/MathML/Values#mathml-specific_types) says they they are encoded as "true" and "false" strings, so for that, we should use BooleanAsTrueFalseStringCodec
- If not sure about a particular attribute, just set the type to string. It's a safe fallback, because all those values are represented as strings in the mathml markup.

If it's not too much work, it would be nice to indicate which element(s) each of these applies to in the scaladoc (the info is in the table)

shared/src/main/scala/com/raquo/domtypes/defs/tags/MathMlTagDefs.scala
```diff
@@ -0,0 +1,334 @@
+package com.raquo.domtypes.defs.attrs
+
+import com.raquo.domtypes.common.{MathMLTagType, TagDef}
+
+
+object MathMLTagDefs {
+
+    val defs: List[TagDef] = List(
```
raquo: Please sort this list alphabetically by tag name (and same with the upcoming attr list – by attr name)

shared/src/main/scala/com/raquo/domtypes/defs/tags/MathMlTagDefs.scala
```diff
@@ -0,0 +1,334 @@
...
+
+      TagDef(
+        tagType = MathMLTagType,
+        scalaName = "math",
+        domName = "math",
+        isVoid = false,
+        scalaJsElementTypeAlias = "org.scalajs.dom.MathMLElement",
```
raquo: Other tag types spell out this alias without the `org.scalajs.` prefix here, so you should remove it here as well.

raquo (commented):
Thanks, good start! 🚀  Please see comments.

.vscode/settings.json
Quafadas: There can be relevant files in that directory, but this particular file I think I agree, should be ignored.

raquo: If that's the case, we should exclude/ignore the entire .vscode diretory, and only specifically include/unignore the files that we do want – if there are any, and if we're really sure that they're needed.

shared/src/main/scala/com/raquo/domtypes/defs/attrs/MathMLDefs.scala
Quafadas: I believe this could be done in a follow up PR? 

I need to read the text carefully, but want to try and get through the tags first, I'd propose to follow this up once this was in a mergable state, in another PR?

jvm/src/test/scala/com/raquo/domtypes/GeneratorSpec.scala
Quafadas: I don't know fi I've captured the full scope of this, but I'm looking

shared/src/main/scala/com/raquo/domtypes/common/TagType.scala
```diff
@@ -16,3 +16,7 @@ object HtmlTagType extends TagType {
 object SvgTagType extends TagType {
   override val sourceStr: String = "SvgTagType"
 }
+
+object MathMLTagType extends TagType {
```
raquo: casing: ML --> Ml

js/src/test/scala/com/thirdparty/defs/tags/MathMLTags.scala
```diff
@@ -0,0 +1,161 @@
...
+
+  /**
+    * The top-level element in MathML, representing a mathematical expression.
+    *
+    * [[https://developer.mozilla.org/en-US/docs/Web/MathML/Element/math math @ MDN]]
+    */
+  lazy val math: MathMlTag[dom.MathMLElement] = mathMLTag("math")
+
+
+  /**
```
raquo: The contents of this file don't match `MathMLTagDefs`, e.g. it does not have the `annotation` element – seems that you need to regenerate again.

raquo (commented):
Sure, we can merge this into a new `mathml` branch (pending a couple new comments).

We can merge this into `master` once the feature is complete and scala-js-dom makes a release.

shared/src/main/scala/com/raquo/domtypes/common/TagType.scala
Quafadas: Hopefully done...

js/src/test/scala/com/thirdparty/defs/tags/MathMLTags.scala
Quafadas: Regenerated... the CI now actually fails on my fork, which hI"m looking into because I don't understand why :-/

Quafadas:
```Run sbt "project domtypesJVM" test "project domtypesJS" +test
/home/runner/work/_temp/6b77182c-5dce-4bb9-b7f4-7b27e2ccac05.sh: line 1: sbt: command not found
```

I'm ... really not sure what's happened.

Quafadas:
Is there any chance I'm hitting this? 

https://github.com/actions/runner-images/issues/10767

Quafadas:
I can't reproduce this failure locally :-(.

raquo:
I've merged this into the new `mathml` branch – as we've discussed, we'll keep MathML work in that branch until it's ready to release.

PR #108, opened 2025-08-04
Continue MathML experimentation

Quafadas:


Quafadas:
This would be the next nibble on https://github.com/raquo/Laminar/issues/172

I had a formatting nightmare on windows, so I setup a devcontainer instead of working locally. that config could be deleted, although it may also be independently useful. 

I _think_ this has a shot at addressing the 4 outstanding points that would be needed in SDT

Quafadas:
Something I think I'm realising, is that not all math ml elements are available on all tags, but this doesn't do anything to attempt to acknowledge that. 

Is that reasonable, or does SDT have some mechanism to say "this attribute is only valid with this tag"?

raquo:
Heyo, sorry I didn't get back to you yet, I'm rather swamped right now. I should be able to review this in the second half of September.

Quafadas:
No problem - there is no rush, but thanks for the message.

raquo:
Thanks for your work @Quafadas!

I have squash-merged the branch of this PR on the command line. [Here it is](https://github.com/raquo/scala-dom-types/commit/783be553565ede497070b0d41003509ace5d1347) in git history – it appears with your authorship properly attached, the same as if I squash-merged this PR. In retrospect I could have just merged this PR and went from there but TBH there was a lot of stuff I needed to do for MathML in Laminar, and I just didn't think this part through. So, I need to close this PR, but your code & commit & authorship is in the next-v19 branch, which will become the master branch soon.

I've been working on finishing MathML support all day today, and it's 99% done. I haven't tested it in the browser yet, only in unit tests. It should work, except that it doesn't support the `class` attribute, or the CSS props yet. I will do those things eventually (a few weeks?), but if you want to play with MathML already, you'll need to:
- get next-v19 branch of Scala DOM Types and +publishLocal it
- get master branches of Airstream and scala-dom-testutils and and +publishLocal it
- get next-v18 branch of Laminar and +publishLocal it (it needs the above)

And then you can use this version of Laminar locally. I wrote a basic MathMlSpec in Laminar, you can start with that code.

This version of Laminar has a bunch of other changes, some of which are binary and source incompatible, but if you just have a [small test project](https://github.com/raquo/scalajs.g8) with no dependencies that depend on com.raquo stuff then you should probably be ok.

If you'd rather wait, I plan to publish v18.0.0-M1 of Laminar in a few weeks, definitely some time this year.

Quafadas:
OOOooohhhhh, you already went all the way to Laminar - wow - very exciting! 

If I can dig out enough time to give it a shot I will...

